PR #4, opened 2017-12-17
Adding constructor and statements + misc

Xapphire13:
- [x] Support byte-order-marker
- [x] Add constructor support
- [x] Expression statements
- [x] Return statements
- [x] Empty statement
- [x] Assignment statements
- [x] Method attributes


Xapphire13:
Sorry about the delay, tests added =]

grammar.js
```diff
@@ -336,9 +359,62 @@ module.exports = grammar({
...
+        '~',
+        '-',
+        '+',
+        '--',
+        '++'
+      ].map(operator => seq(operator, $._expression))
+      .concat([
+        'typeof',
+        'sizeof'
+      ].map(operator => seq(operator, '(', $._expression, ')'))))),
```
maxbrunsfeld: This will mean that the `unary_expression` node will have a varying number of children. Usually, it will just have two children (the operator and the operand expression), but in the case of `typeof` and `sizeof`, it will have four children (operator, left-paren, operand expression, and right-paren).

It might make the tree a little more uniform and easy to analyze if we treated `typeof` and `sizeof` just like any other unary operator. We could handle the parentheses independently via a `parenthesized_expression` rule which we'll eventually need anyway.

```js
parenthesized_expression: $ => seq('(', $._expression, ')'),
```

Xapphire13: That makes sense, and parenthesis are always legal so this doesn't break anything =]

maxbrunsfeld: If we do want to enforce that parentheses are used with `sizeof` and `typeof`, we could keep your basic structure, but replace `seq(operator, '(', $._expression, ')')` with `seq(operator, $.parenthesized_expression)`.

I'd be ok with either approach. In general, I'm ok w/ parsing a *superset* of the language in question (e.g. allowing things like `typeof foo`) if it makes the syntax tree or the grammar implementation simpler.

Xapphire13: I like going for the superset approach, keeps things simple. And if this becomes an issue, someone will open a bug, and we already know the solution.

maxbrunsfeld:
Thanks @Xapphire13, Great progress!

PR #27, opened 2019-11-19
Await expression needs to be sequence not choice

initram:
I am not sure if I also have to commit the changes to the auto generated files, or why they are even checked in at all.

A guide for contributing to this repo would go a long way.


damieng:
Thanks!

I've been working from the documentation at https://tree-sitter.github.io/tree-sitter/creating-parsers - if there is something additional you think this repo could use let me know.

PR #28, opened 2019-11-19
Remove auto-generated files from source control

initram:
They will now be generated during `npm test`, which is run on the build server.

This should make working with the repo more strait forward. No need to commit generated files and having huge diffs that are hard to review.


initram:
hmm... This is not as simple as I first thought. Why it feels the need to build the generated code during `npm install` is beyond me. I see that the other tree-sitter projects also commit the generated source, so I guess that is just how it is done for tree-sitter grammas.

This is most likely something that should be delt with on the main tree-sitter repo, so I will close this.

initram:
I have now tried to remove the last auto generated files and added `node-gyp configure` to the npm build script, so that it should still be possible to build the final assembly, but it does not try to do it when running `npm install` as there is no bindings.gyp file at that point.

damieng:
Thanks for investigating this but right now it's advantageous to keep this repo as well-aligned with the layout and process of all the other tree-sitter grammars when it comes to the package.json and checked-in generated files.

e.g.

- https://github.com/tree-sitter/tree-sitter-rust
- https://github.com/tree-sitter/tree-sitter-c
- https://github.com/tree-sitter/tree-sitter-python

damieng:
Additionally I just found out that some projects are including tree-sitters as submodules right now so this would break them.

PR #29, opened 2019-11-20
Added the array creation grammar to expression

initram:
I also changed the grammar of implicit array creation to use array_initalizer, as it did not support multi dimentional arrays without it.

I had to remove "new" from modifier rule to make this work, but the modifier rule should really be split into multiple rules as not all modifiers are allowed everywhere. See e.g.: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#class-modifiers


maxbrunsfeld:
Thank for writing good tests.

Regarding the modifier rules: In an LR parser, trying to split out the different modifier rules introduces an untenable number of ambiguities (since you don't know what kind of declaration you're parsing until after you're a few tokens *past* the modifiers). If you can come up with a way to do it that doesn't require adding more conflicts than it removes, I'm open to it, but as far as I understand, the current approach is going to work better.

PR #31, opened 2019-11-20
added precompiler directive #nullable

initram:
This is new in C# 8, and used by Newtonsoft.Json files that the scripts test againts.

Should I also commit the changes to known-failures.txt, or should i leave that alone?


damieng:
Thanks!

damieng:
@maxbrunsfeld What's the process around known-failures.txt ?

PR #38, opened 2019-11-21
added @identifiers

initram:
They are used when you want to use a keyword name as an identifier name.

I did not add the @ in the regex as it is not part of the name of the identifier. `@x == x` this would be comparing a variable with it self. Not sure how to test this, or if it is needed at all.


package.json
```diff
@@ -16,6 +16,6 @@
     "tree-sitter-cli": "^0.15.14"
   },
   "scripts": {
-    "test": "tree-sitter test"
+    "test": "tree-sitter generate && tree-sitter test"
```
maxbrunsfeld: I don't think we want to regenerate every time we run tests, because regenerating and recompiling takes much longer. Could you back this part out?

initram: Yes that was a mistake! I use it like that locally to verify my changes to the grammar.

damieng:
@initram When you create the PR can you allow maintainers to push to your branch?  It's easier for us to resolve conflicts there so we can squash and merge PRs. Thanks!

damieng:
Reduces known-failures by another 11 files.

PR #39, opened 2019-11-21
Added out var and var as implicit_type

initram:
I hope it is okay that i added two changes in one PR.


damieng:
Yeah, that's fine especially when they're as closely related.  Thanks!

PR #43, opened 2019-11-22
Allow qualified names in attributes

initram:


damieng:
Further reduces the number of failing files by 11 bringing the file-level success rate to 92.6%!

PR #44, opened 2019-11-24
Add query_expression syntax

### LINKED ISSUE #22: Enable query_expression without conflicts
Would be nice to have the LINQ comprehension syntax working however enabling `query_expression` by un-commenting it in `_expression` causes lots of conflicts.  It needs some precedence specifying.

damieng:
Fixes #22

Reduces known-failures by another 12 bringing test-suit up to 93.7% files parsed ok.

Also introduces the `_reserved_identifiers`  which will need to be fleshed out.


initram:
From is known as a contextual keyword in C#, so it is not a reserved word. See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/#contextual-keywords

I think the lexer will need to be able to provide an ambiguous token to the parser that can then be solved there.

maxbrunsfeld:
The way that we deal with contextual keywords in other languages is to define a rule called something like ‘reserved_identifier’ that matches those keywords, aliased as an ‘identifier’. Then, in certain rules that accept identifier, we make them alternatively accept a ‘reserved_identifier’. Here’s an example from the JavaScript grammar, in which ‘get’ and ‘set’ are contextual keywords:

https://github.com/tree-sitter/tree-sitter-javascript/blob/master/grammar.js#L1049

damieng:
Cool, will check that out tonight.

Issue #24, opened 2019-11-18
Enable ref_expression without conflicts


damieng:
Enabling `ref_expression` causes conflicts with the `ref` modifier.  Some resolution is needed.

PR #45, opened 2019-11-25
Added ref_expression

initram:
I had to increase the precedence of argument rule to disambiguate the case where an argument is passed by reference.

Please let me know if there is a better way to specify that we prefer `argument` over `ref_expression`


initram:
This would solve #24

PR #49, opened 2019-11-26
Wrap up all but one reserved identifier (global)

### LINKED ISSUE #47: Complete contextual keywords
There are a [number of keywords in C# that are contextual](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/#contextual-keywords).

They should be added to `_reserved_identifiers` and appropriate conflict rules setup between _reserved_identifiers and the rule that implements the matching for when it is used as a keyword (see `from` for example).

damieng:
Fixes #47

Some are disabled right now as they cause regressions with the examples.

### DIFF
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -1,8 +1,65 @@
 =======================================
-From keyword can be a variable
+Add keyword is contextual
 =======================================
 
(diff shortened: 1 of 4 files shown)


damieng:
Merged manually.

PR #51, opened 2019-11-27
Fix for member access on predefined types

initram:
This was not working previously as _type rule was commented out in expression rule.

There are 3 additions to known-failures.txt, but those should be fixed with #50.


damieng:
Awesome, thanks!  That rule has been a pain.

damieng:
Coverage on the example projects is up to 97.6% of the files!

PR #53, opened 2019-11-28
Fixed property_declaration rule

initram:
The explicit_interface_specifier rule should be after the _type rule.


corpus/type-properties.txt
```diff
@@ -143,3 +143,24 @@ class Foo {
         (identifier_name)
         (accessor_declaration) (accessor_declaration)
         (integer_literal)))))
+
+=====================================
+Class with expplicit interface properties
```
damieng: Extra `p` seems to have snuck in here.

PR #62, opened 2019-12-03
Tuple_type added

### LINKED ISSUE #15: Add tuple_type
The `_type` in grammar.js does not yet include [tuple_type](https://github.com/tree-sitter/tree-sitter-c-sharp/blob/master/grammar.txt#L375) because it causes multiple conflict rules and issues.

initram:
I had to add some conflicts as expected, and fixed a small error in the rule, but other than that it was not bad.

I think we should work on the grammar for deconstruction, as it currently parses it to something a bit weird (depending on the syntax used). See https://docs.microsoft.com/en-us/dotnet/csharp/deconstruct

Fixes #15

### DIFF
--- a/corpus/statements.txt
+++ b/corpus/statements.txt
@@ -1014,3 +1014,58 @@ class A {
       (block
         (for_statement
           (block)))))))
+
+=====================================
+Deconstruction
+=====================================
+
+class A {
+  void Sample() {
+    (var a, var b) = c;
+    var (a, b) = c;
+    (a, b) = c;
+    var (a, _) = c;
+  }
+}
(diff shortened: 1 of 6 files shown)


damieng:
Thanks!

maxbrunsfeld:
Thanks @initram! In https://github.com/tree-sitter/tree-sitter-c-sharp/commit/57ba56f25b415da5b3dd2e7381387fa99d71b314, I did some work to try and detangle the expression/type parsing. The tuples and deconstructing patterns in the test that you added are now parsing more according to expectation, I think.

PR #64, opened 2019-12-13
Fix path to binding

rien:
The filename of the binding has changed, resulting in an error when trying to import `tree-sitter-c-sharp`:
```
  Error {
    code: 'MODULE_NOT_FOUND',
    message: 'Cannot find module \'./build/Release/tree_sitter_c_sharp_binding\'',
  }
```

I have renamed the path in `index.js` and have confirmed this fixes the issue.

I also propose adding an integration test which tries to import the `tree-sitter` grammar(s) in JavaScript to prevent bugs like this from happening.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,7 +1,7 @@
 {
   "targets": [
     {
-      "target_name": "ts_language_c_sharp_binding",
+      "target_name": "tree_sitter_c_sharp_binding",
       "include_dirs": [
         "<!(node -e \"require('nan')\")",
         "src"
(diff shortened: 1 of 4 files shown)


rien:
Sure! I'll look into both things later this weekend.

rien:
@maxbrunsfeld done!

maxbrunsfeld:
Thanks!

rien:
@maxbrunsfeld 0.16.1 doesn't seem to be published yet, can you push the release to npm?

maxbrunsfeld:
[Npm](https://www.npmjs.com/package/tree-sitter-c-sharp) is showing that it was published 17 hours ago. I think things just take while to "propagate" across npm's servers? 🤷‍♂

rien:
Hmm, yeah I see it now. Weird.

Thanks anyway :smile:

PR #91, opened 2020-09-25
Changed initializer_expression to not be a top level expression

initram:
Corrected test that can not compile in real C# compiler. Also changed initializer_expression slightly as it is no longer a direct child of _expression.
This also allows us to remove the conflict with block rule.

I was working on property patterns and ran into this. It was causing conflicts with the rules I were setting up, so this is also a step on the way for property patterns. But I thought that I would split it into smaller pull requests, as they are easier to discuss.

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -200,12 +200,13 @@ void b() {
               (anonymous_object_creation_expression (identifier)))))))))
 
 ============================
-Anonymous object creation with single named
+Anonymous object creation with two named
 ============================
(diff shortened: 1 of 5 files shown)


initram:
What do you think of these changes? It makes some of the rules more complex, but it allows us to avoid some conflicts. I also think that the resulting syntax tree looks more correct for some cases of initializers now. But I have not compared with roslyn outout.

initram:
This was not needed anyway

PR #98, opened 2020-09-30
Added the recursive_pattern rule

### LINKED ISSUE #19: Enable recursive_pattern rule (for positional and property patterns)
The Roslyn grammar.txt defines `recursive_pattern` as:

```antlr
recursive_pattern
  : type? positional_pattern_clause? property_pattern_clause? variable_designation?
```

Which means everything is optional.  You can't do that in tree-sitter so need an alternative.

initram:
This will parse the most common uses of the recursive patterns, while it still fails or parses some others wrong.

fixes #19 

The following is still parsed wrong:
// this is parsed as a (constant_pattern(invocation_expression ...)), should be (recursive_pattern .. (positional_pattern_clause ...))
var b = p is Point (var x, var y); 

//this causes a parse error as it tries to parse it as invocation_expression again
var b = p is Point (var x, var y) { x: 0 };

### DIFF
--- a/Martin todo.txt
+++ b/Martin todo.txt
@@ -1,0 +1,13 @@
+there is some implementation of "recursive patterns" that looks like positional patterns.
+
+allign this with the property patterns i made and remove dublicate rules.
+
+
+see gramma here: https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/patterns.md
+
+
+Known issues:
+
+var b = p is Point (var x, var y); // is parsed as a (constant_pattern(invocation_expression ...)), should be (recursive_pattern .. (positional_pattern_clause ...))
+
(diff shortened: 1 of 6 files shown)


initram:
What do you think? should we take the changes as they are now, or should we wait until I have gotten it to parse more of the recursive patterns?

Also do I need to add more comments to the `prec(1, ...)` I have added?

damieng:
I think incremental work over perfect is always better - it's pretty much how this grammar has evolved so I have no problem with this if it's not causing any regressions.

Any thoughts @maxbrunsfeld or @patrickt ?

damieng:
In the absence of any additional feedback I say let's take this. @initram do you want to fix up the conflicts?

initram:
Sure, I will fix the conflicts.

initram:
There should be no more conflicts now.

damieng:
Thanks!

PR #105, opened 2020-10-05
Added support for lambda discard parameters

initram:
I don't know if the discard should be part of the field called 'name' or not, as it is really not a name of an identifier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Comprehensive support for C# exists with the following exceptions:
 - [ ] Extension GetEnumerator
 - [ ] Function pointers
 - [x] Init properties
-- [ ] Lambda discard parameters
+- [x] Lambda discard parameters
 - [ ] Local function attributes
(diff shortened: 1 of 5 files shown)


damieng:
Hmm. They're just treated as identifiers in Roslyn:

image

image

That said, I'm not opposed to giving them more semantic meaning if you can see value.

initram:
huh, I just reused the discard token we had used in other places. It seems that roslyn adds the discard logic on in a later pass or something.
`int.TryParse("1", out _)`
Here roslyn also syntactically treats _ as an identifier, but with some extra metadata that says it is discard. Where we say that it is a discard token.

I think it makes sense what we do, we are not trying to create a copy of roslyn anyways.

damieng:
Hi, I tried emailing you on gmail but no answer. Seems we're both hacking on this repo today and I want to make sure we don't overlap. I'm mostly done on C# 9 records.

PR #107, opened 2020-10-05
Support records + with keyword

### LINKED ISSUE #106: Implement C# 9 record support
This should cover the `record_declaration` type which is similar but not identical to `class_declaration` as well as the difference in bases to cover the argument passing.

Additionally should include `with` keyword support.

damieng:
Fixes #106 however...

The proposal seems unpolished (I found an error in the syntax definitions - reported) and the Roslyn compiler generates differently named expressions to that in the proposal (unusual).

I've gone with the Roslyn definitions however I suspect from the naming that you can do more with the `with` keyword than the proposal really touches on so that may need further rules later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Comprehensive support for C# exists with the following exceptions:
   - [ ] Simple type patterns
   - [x] Relational patterns
   - [x] Logical patterns
-- [ ] Records
+- [x] Records
 - [ ] Skip Locals Init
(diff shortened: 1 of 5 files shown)


initram:
Looks good. 

I think a test that shows usage of the record body would make sense to have.

Also I think the left hand side of a with expression should be _expression as any expression that results in a record type can be used with a with-expression, e.g., a function invocation that returns a record.

PR #112, opened 2020-10-06
Added function pointer support

### LINKED ISSUE #111: Add support for function pointers
I will try to implement this.

initram:
Fixes #111

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -581,6 +581,7 @@ module.exports = grammar({
       $._name,
       $.nullable_type,
       $.pointer_type,
+      $.function_pointer_type,
       $.predefined_type,
       $.tuple_type,  // TODO: Conflicts with everything
     ),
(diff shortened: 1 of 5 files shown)


grammar.js
```diff
@@ -608,6 +609,36 @@ module.exports = grammar({
...
+
+    unmanaged_calling_convention: $ => choice(
+      'Cdecl',
+      'Stdcall',
+      'Thiscall',
+      'Fastcall',
+      commaSep1($.identifier)
+    ),
+
+    funcptr_parameter: $ => seq(
```
damieng, line 637: Roslyn calls this `FunctionPointerParameter` so perhaps `function_pointer_parameter` ?

grammar.js
```diff
@@ -608,6 +609,36 @@ module.exports = grammar({
...
+    function_pointer_type: $ => seq(
+      'delegate',
+      '*',
+      optional($.calling_convention_specifier),
+      '<',
+      commaSep1($.funcptr_parameter),
+      '>'
+    ),
+
+    calling_convention_specifier: $ => choice(
```
damieng, line 621: Roslyn calls this `FunctionPointerCallingConvention` so perhaps `function_pointer_calling_convention` ?

grammar.js
```diff
@@ -608,6 +609,36 @@ module.exports = grammar({
...
+      'managed',
+      seq(
+        'unmanaged',
+        optional(seq('[', $.unmanaged_calling_convention, ']'))
+      )
+    ),
+
+    unmanaged_calling_convention: $ => choice(
```
damieng, line 629: Roslyn definitely seems to vary from the proposal grammar on this one. They have line 625 as a rule called `FunctionPointerUnamangedCallingConventionList` and then each convention inside is a `FunctionPointerUnmanagedCallingConvention`.

grammar.js
```diff
@@ -608,6 +609,36 @@ module.exports = grammar({
...
+      'Cdecl',
+      'Stdcall',
+      'Thiscall',
+      'Fastcall',
+      commaSep1($.identifier)
+    ),
+
+    funcptr_parameter: $ => seq(
+      repeat(choice('ref', 'out', 'in', 'readonly')),
```
damieng, line 638: It can only be one of these, perhaps add `ref readonly` as the final option instead?

damieng (changes requested):

grammar.js
damieng: Also think we're missing corpus coverage for this.

initram:
I have now made changes to align it more with the AST produced by roslyn. Let me know if there are any other feedback.

damieng (approved):

damieng:
Looks great, thanks!

PR #114, opened 2020-10-06
Support top level statements from C# 9

### LINKED ISSUE #113: Implement top level statement support from C# 9.0
This will cause a large change in the corpus as previously we deviated from Roslyn to allow partial fragments of code to be syntax highlighted and parsed... and we did that by allowing all sorts of declarations at the top level.

Now there is a specific syntax and tree for it a lot of previously accepted-top level items will be treated differently. I expect we'll still support them all but there will be subtle changes like:

```csharp
var a = Assert.Range(from, to);
```

On it's own will change from being a `field_declaration` (we treated it as if it were within a class even tho `var` isn't valid on a field declaration) to a `local_declaration_statement` just as if it were in a function.

damieng:
Now that C# 9.0 supports top-level statements we can remove our workarounds we were using for fragment markup.

This means code we previously hacked as `field_declaration` is now `local_declaration_statement` and `method_declaration` is `local_function_statement` (basically we now treat functions and variables as if they were defined inside an invisible function like C# 9.0 specifies - previously we treated them as if they were on an invisible class).

Fixes #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Comprehensive support for C# exists with the following exceptions:
 - [x] Static anonymous functions
 - [ ] Target-typed conditional expression
 - [x] Target-typed new
-- [ ] Top-level statements
+- [x] Top-level statements
 
 ### References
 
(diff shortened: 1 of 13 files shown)


damieng:
Fixes #113

Issue #12, opened 2019-11-18
Add global_statement


damieng:
The `_member_declaration` in grammar.js does not yet include:

- [global_statement](https://github.com/tree-sitter/tree-sitter-c-sharp/blob/master/grammar.txt#L309)

It causes all sorts of conflicts/precedence situations that would need resolving.

initram:
I think we should just close this case now. I don't think there is any more work to be done here.

PR #115, opened 2020-10-06
Treat top-level as global statements as per Roslyn

damieng:
Treats newly added top-level statements as global statements.

Should tie in with #12 and see what syntax is still left to support there.

This is about as far as we can go with global statements without seeing the grammar for how Roslyn has repurposed them etc.

### DIFF
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -7,12 +7,13 @@ var a = Assert.Range(from, to);
 ---
 
 (compilation_unit
-  (local_declaration_statement
-    (variable_declaration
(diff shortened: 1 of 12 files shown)


initram:
I think this looks fine. As far as I can see in the tests in the Roslyn repo, it is only used for statements declared at the top level like you use it here. I do not see how it was supposed to be used before c# 9, and the gramma.txt file does not really help much. Maybe it was used as a form of error recovery, but who knows.

damieng:
Yeah it's only missing attributes and modifiers and I can't see how they were ever used.

Issue #131, opened 2020-10-09
Conditional access pattern failure


damieng:
This kind of `if` pattern is not working in the case of 

```csharp
if ((me as Person)?._Age != 1)
{
}
```

This looks like it could be to do with the priority of `conditional_access_expression`.

initram:
I think there are multiple things that are going wrong here. One of them is that we allow real_literals to start with under-score before a number, changing the rules for integer_literal and real_literal helps a bit. Here is the code I have tried:
`    integer_literal: $ => token(seq(
      choice(
        (/[0-9][0-9_]*/), // Decimal
        (/0[xX][0-9a-fA-F_]+/), // Hex
        (/0[bB][01_]+/) // Binary
      ),
      optional(/u|U|l|L|ul|UL|uL|Ul|lu|LU|Lu|lU/)
    )),

    null_literal: $ => 'null',

    real_literal: $ => {
      const suffix = /[fFdDmM]/;
      const exponent = /[eE][+-]?[0-9_]+/;
      return token(choice(
        seq(
          (/[0-9][0-9_]*/),
          '.',
          (/[0-9][0-9_]*/),
          optional(exponent),
          optional(suffix)
        ),
        seq(
          '.',
          (/[0-9][0-9_]*/),
          optional(exponent),
          optional(suffix)
        ),
        seq(
          (/[0-9][0-9_]*/),
          exponent,
          optional(suffix)
        ),
        seq(
          (/[0-9][0-9_]*/),
          suffix
        )
      ))
    },`

Then it should parse, but the precedence is wrong. It interpreted it kind of like this:

`if ((me as Person)?((._Age) != 1))
{
}`

damieng:
I've put up a fix for the literals at #132 - it's a bit different as it turns out that literals should also not end with `_` either.

initram:
we should also make sure that the precedence is right for `var c = (char)(me as string)?[1];`, `var c = !(me as Person)?.IsOld;`, `var c = await (me as Person)?.SomethingAsync();`

I am currently trying to solve some of these issues, but I may create a pull request that only fixes a subset to begin with.

damieng:
I've been hacking on this as well - there's definitely more than one problem here - even regular `conditional_expressions` are incorrect, e.g.

```csharp
var a = b ? 1 + 2 : 3;
```

Is generating:
```
(equals_value_clause
  (conditional_expression
    (identifier)
[...]

PR #132, opened 2020-10-10
Underscore not permitted at start or end of numeric literal segment

damieng:
When `_` numeric literal support was added it was permitted everywhere in the numeric/hex/boolean literal.

It should not be allowed at the start of end of a segment.

We were also missing a test case for real literals starting with a `.`

Discovered in #131

### DIFF
--- a/corpus/literals.txt
+++ b/corpus/literals.txt
@@ -5,9 +5,11 @@ integer literals
 const int dec = 1_2;
 const long hex = 0xf_1l;
 const long hex2 = 0Xffff;
+const long hex3 = 0x_0_f;
 const UInt64 dec = 1uL;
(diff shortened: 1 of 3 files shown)


initram:
Actually this is allowed 0x_12 and 0b_1010

But otherwise it looks good.

damieng:
Ah, the joys of them not bothering to actually update a proper spec but just throwing out English language interpretations.

Will fix and push again.

initram:
I think you now allow binary and hex to end in underscore, which I don't think is allowed. I think you can just swap the star and plus in those regexes. Otherwise test them using regex101.com

initram (approved):
Looks good to me

PR #139, opened 2020-11-10
Constrain conditional access to avoid conflicts etc

### LINKED ISSUE #131: Conditional access pattern failure

damieng:
Changes the allowed type of conditional access from the open-ended `expression` in the Roslyn grammar to simply allow `?.` and `?[x]` which is more inline with the C# 6 specification grammar.

Does not introduce any new regressions in our examples parsing and drops the Roslyn grammar parsing failures from 27 files to 25 (most of which are caused by statement splitting in preprocessor blocks)

Fixes #131 and #138.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1086,10 +1086,10 @@ module.exports = grammar({
       seq('unchecked', '(', $._expression, ')')
     ),
 
-    conditional_access_expression: $ => prec.right(seq(
+    conditional_access_expression: $ => prec.right(PREC.COND, seq(
       field('condition', $._expression),
       '?',
-      field('value', $._expression)
+      choice($.member_binding_expression, $.element_binding_expression)
     )),
 
     conditional_expression: $ => prec.right(PREC.COND, seq(
@@ -1206,7 +1206,7 @@ module.exports = grammar({
 
     member_binding_expression: $ => seq(
       '.',
-      $._simple_name,
+      field('name', $._simple_name),
     ),
 
     object_creation_expression: $ => prec.right(seq(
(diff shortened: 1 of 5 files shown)


grammar.js
```diff
@@ -1086,10 +1086,10 @@ module.exports = grammar({
       seq('unchecked', '(', $._expression, ')')
     ),
 
-    conditional_access_expression: $ => prec.right(seq(
+    conditional_access_expression: $ => prec.right(PREC.COND, seq(
       field('condition', $._expression),
       '?',
-      field('value', $._expression)
+      choice($.member_binding_expression, $.element_binding_expression)
```
maxbrunsfeld, line 1092: Are `member_binding_expression` and `element_binding_expression` used anywhere else still? If I understand them correctly, I think that they can *only* appear here in this conditional access expression.

damieng: They're both still listed under options in `expression` as per the Roslyn syntax.

We need element_binding_expression there for syntax like

```csharp
var x = new Dictionary<string, int> { ["a"] = 65 };
```

I believe `member_binding_expression` can be removed from `expression` although it does cause the number of file regressions in Roslyn to go up by 1 but that is purely because of the known restriction we currently have with `await` not being able to be used as an identifier and the line

```csharp
isDynamic = await.AwaitableInfo.IsDynamic;
```

Parsing as a side-effect of allowing unqualified access to a member (it's treating it as `await .AwaitableInfo.IsDynamic`

damieng: I've pushed an update that removes `member_binding_expression` from `expression`.

PR #145, opened 2021-01-08
Change precedence of select in LINQ queries

### LINKED ISSUE #144: Incorrect precedence in LINQ query
I have the following C# code:

```csharp
            var numQuery =
                from num in numbers
                select num * 2;
```

This is parsed by tree sitter as:
```csharp
            var numQuery =
                (from num in numbers
                select num) * 2;
```

I expect it to be parsed like this:
```csharp
            var numQuery =
                from num in numbers
                select (num * 2);
```

Here is the tree tree-sitter makes:
```
              (local_declaration_statement [11, 12] - [15, 31]
                (variable_declaration [11, 12] - [15, 30]
                  type: (implicit_type [11, 12] - [11, 15])
                  (variable_declarator [11, 16] - [15, 30]
                    (identifier [11, 16] - [11, 24])
                    (equals_value_clause [11, 25] - [15, 30]
                      (binary_expression [12, 16] - [15, 30]
                        left: (query_expression [12, 16] - [15, 26]
                          (from_clause [12, 16] - [12, 35]
                            (identifier [12, 21] - [12, 24])
                            (identifier [12, 28] - [12, 35]))
                          (comment [13, 16] - [14, 41])
                          (select_clause [15, 16] - [15, 26]
                            (identifier [15, 23] - [15, 26])))
                        right: (integer_literal [15, 29] - [15, 30]))))))
```

Sjord:
In a LINQ query, e.g. `from w in words select w`, anything after the `select`
should be included in the query. So `select w * 2` should be parsed as `select
(w * 2)`.

I noted this problem with multiplication and function calls. I also added tests
for conditional comparison and assignment, because these normally have the
highest precendence.

I wrote the tests, and then changed things in grammar.js till it worked. So
this change comes from result-based effort, not a good understanding of
operator precedence.

Tested with `tree-sitter generate && tree-sitter test`.

Fixes #144

### DIFF
--- a/corpus/query-syntax.txt
+++ b/corpus/query-syntax.txt
@@ -24,6 +24,110 @@ var x = from a in source select a.B;
                     (identifier))))))))))
 
 =====================================
+Query from select with operator
+=====================================
+
+var x = from a in source select a * 2;
+
+---
+
+(compilation_unit
+  (global_statement
+    (local_declaration_statement
+      (variable_declaration
+        (implicit_type)
+        (variable_declarator
+          (identifier)
+          (equals_value_clause
+            (query_expression
+              (from_clause
+                (identifier)
+                (identifier))
+              (select_clause
+                (binary_expression
+                  (identifier)
+                  (integer_literal))))))))))
+
+=====================================
+Query from select with method call
+=====================================
+
+var x = from a in source select a.B();
+
+---
+
+(compilation_unit
+  (global_statement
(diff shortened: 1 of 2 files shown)


damieng (commented):
Looks good, thanks!

Yes, the precedence thing is a bit of trial and error sometimes but as long as all the tests pass we're good :)

Issue #149, opened 2021-02-09
Incorrect precedence with coalesce ?? operator and equals ==


Sjord:
```csharp
var b = obj ?? a == 0;
```

Actual: `(obj ?? a) == 0`
Expected: `obj ?? (a == 0)`

damieng:
Fixed in #156

PR #156, opened 2021-02-10
Fix `??` priority and give it a name

damieng:
`??` produced confused expression trees in #149

The priority was incorrect.

Also took the opportunity to introduce `coalesce_expression` as per Roslyn.

```csharp
b = obj ?? a == 0;
```

Results in

```
assignment_expression
  identifier
  assignment_operator
  coalesce_expression
    identifier
    binary_expression
      identifier
      integer_literal
```

Introduces no regressions in the corpus tests or the example suite.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1397,7 +1397,6 @@ module.exports = grammar({
       $.binary_expression,
       $.cast_expression,
       $.checked_expression,
-      $.coalesce_expression,
       $.conditional_access_expression,
       $.conditional_expression,
       $.default_expression,
@@ -1457,6 +1456,7 @@ module.exports = grammar({
         ['!=', PREC.EQUAL],
         ['>=', PREC.REL],
         ['>', PREC.REL],
+        ['??', PREC.TERNARY],
       ].map(([operator, precedence]) =>
         prec.left(precedence, seq(
           field('left', $._expression),
@@ -1465,12 +1465,6 @@ module.exports = grammar({
         ))
       )
     ),
-
-    coalesce_expression: $ => prec.left(PREC.TERNARY, seq(
-      field('left', $._expression),
-      field('operator', '??'),
-      field('right', $._expression)
-    )),
 
     as_expression: $ => prec.left(PREC.EQUAL, seq(
       field('left', $._expression),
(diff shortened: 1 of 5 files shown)


grammar.js
```diff
@@ -1396,6 +1397,7 @@ module.exports = grammar({
       $.binary_expression,
       $.cast_expression,
       $.checked_expression,
+      $.coalesce_expression,
```
maxbrunsfeld, line 1400: Adding more rules does increase the size of the parse table and the API surface area slightly. So oftentimes, if there's a set of language constructs that are all valid in the same contexts, and have the same internal structure, we'll try to describe them all with the same rule. Of course, it's fine to separate them out if there's some distinct semantic behavior associated with one of the constructs.

In this case, I would think that `??` "belongs" together with other short-circuiting binary operations like `||` and `&&`. I don't know of any downstream code that needs to treat them differently. Do you think there's a good reason to have a dedicated rule for it?

damieng: It's just for compatibility with the Roslyn grammar. Although I know we don't aim to reproduce it keeping it close makes it a lot easier to debug and maintain especially as C# has been evolving so much over the last few years. (C# 7.0, 7.1, 7.2, 8.0 and 9.0 since 2017)

damieng: That said, in this particular case I could go either way - would probably just add 'coalesce_expression' as a comment onto the binary_expression it relates to (and perhaps flesh out the others)

maxbrunsfeld: I think that since the roslyn grammar is inherently so different from ours (it's more of a loose description of the trees that roslyn creates, as opposed to a formal grammar that's used to generate their parser), a 1:1 match shouldn't necessarily be a goal. It is good to be informed by their naming, but I think it's not as high priority as our concrete desires to reduce the API complexity and the code size of the grammar. So for this PR, I think my preference would be to fix the precedence, but keep it as `binary_expression`.

damieng: Is there a way of putting the operators into the corpus tests? Would help with making them clearer when everything is a binary_expression.

maxbrunsfeld: Yeah, I see the problem you're talking about - trying to assert about the nesting level. There isn't a way to put them in the tests right now, but you can test it by making the *operands* distinctive :

```
b = null ?? a == 0
```

This way, the test will only pass if the precedences are correct.

maxbrunsfeld (approved):

Issue #152, opened 2021-02-10
Incorrect precedence between unary prefix operator and element access expression


Sjord:
```csharp
var b = +a[0];
```

Expected: `+(a[0])`
Actual: `(+a)[0]`

damieng:
Fixed in 645098a1e6c1aeb1cd6dc8bf0af44970a25790c3

PR #157, opened 2021-02-10
Fix prec for element access vs cast

### LINKED ISSUE #147: Incorrect precedence with cast and array access
```csharp
int[] a = { 0 };
var b = (float)a[0];
```

The `(float)a[0]` part gets incorrectly parsed as `((float)a)[0]`, and I expect it to be parsed as `(float)(a[0])`.

damieng:
Fixes #147


maxbrunsfeld (approved):

damieng:
Also fixes #152, will check in corpus coverage separately

Issue #166, opened 2021-03-07
Array of named tuples incorrectly parsed as element_access_expression instead of array_creation_expression


Sjord:
```
var greetings = new (string head, string tail)[3];
```

Actual:

          (local_declaration_statement [7, 8] - [7, 58]
            (variable_declaration [7, 8] - [7, 57]
              type: (implicit_type [7, 8] - [7, 11])
              (variable_declarator [7, 12] - [7, 57]
                (identifier [7, 12] - [7, 21])
                (equals_value_clause [7, 22] - [7, 57]
                  (element_access_expression [7, 24] - [7, 57]
                    expression: (implicit_object_creation_expression [7, 24] - [7, 54]
                      (argument_list [7, 28] - [7, 54]
                        (argument [7, 29] - [7, 40]
                          (declaration_expression [7, 29] - [7, 40]
                            type: (predefined_type [7, 29] - [7, 35])
                            name: (identifier [7, 36] - [7, 40])))
                        (argument [7, 42] - [7, 53]
                          (declaration_expression [7, 42] - [7, 53]
                            type: (predefined_type [7, 42] - [7, 48])
                            name: (identifier [7, 49] - [7, 53])))))
                    subscript: (bracketed_argument_list [7, 54] - [7, 57]
                      (argument [7, 55] - [7, 56]
                        (integer_literal [7, 55] - [7, 56])))))))))))))

Expected:

          (local_declaration_statement [7, 8] - [7, 48]
            (variable_declaration [7, 8] - [7, 47]
              type: (implicit_type [7, 8] - [7, 11])
              (variable_declarator [7, 12] - [7, 47]
                (identifier [7, 12] - [7, 21])
                (equals_value_clause [7, 22] - [7, 47]
                  (array_creation_expression [7, 24] - [7, 47]
                    (array_type [7, 28] - [7, 47]
                      type: (tuple_type [7, 28] - [7, 44]
                        (tuple_element [7, 29] - [7, 35]
                          type: (predefined_type [7, 29] - [7, 35]))
                          name: (identifier [?, ?] - [?, ?])))
                        (tuple_element [7, 37] - [7, 43]
                          type: (predefined_type [7, 37] - [7, 43])))
                          name: (identifier [?, ?] - [?, ?])))
                      rank: (array_rank_specifier [7, 44] - [7, 47]
                        (integer_literal [7, 45] - [7, 46])))))))))))))

Sjord:
[...]

PR #167, opened 2021-03-21
Parse arrays of named tuples correctly

Sjord:
`new (string a, string b)[3]` can be parsed as element access on a implicit
object creation, or as a array creation expression of a named tuple. Give
precedence to the latter.

Closes issue #166.


damieng:
Thank-you for this fix!

PR #174, opened 2021-03-25
Null conditional operations

Sjord:
Implement conditional access expressions according to [the C# lang specification](https://github.com/dotnet/csharplang/blob/main/spec/expressions.md#null-conditional-operator).

I am not totally convinced this is better than [the way Roslyn does it](https://github.com/tree-sitter/tree-sitter-c-sharp/issues/138#issuecomment-806473913), but this gives something working to compare against.

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -1949,12 +1949,17 @@ if ((p as Person[])?[0]._Age != 1) { }
   (global_statement
     (if_statement
       (binary_expression
-          (member_access_expression
-            (conditional_access_expression
-              (parenthesized_expression
-                (as_expression (identifier) (array_type (identifier) (array_rank_specifier))))
-              (null_conditional_element_access
-                (argument (integer_literal))))
-            (identifier))
-          (integer_literal))
+        (conditional_access_expression
+          (parenthesized_expression
+            (as_expression
+              (identifier)
+              (array_type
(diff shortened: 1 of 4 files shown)


damieng:
Thanks for tackling this, a couple of immediate things come to mind.

1. Be wary of relying too much on the C# language specifications you find online. None of them have been updated post C# 5.0 and so are superseded in many area by fragments in C# feature proposals that were adopted in 6.0-9.0. MadsTorgersen did say they were working on it but I haven't seen any sign of it yet and that was at least 6 months ago :( This is why I've used Roslyn as the basis for the grammar rather than trying to piece together and reconcile fragments.

2. You'll need to change `'?.'` and `'?['` to `seq('?','.')` and `seq('?', '[')` as the language allows white-space between them;

image

damieng:
Will take a look at this one tomorrow :)  Also they're finally working on ratifying all the passed proposals back into a single spec now https://github.com/dotnet/csharpstandard is where they are at so far.

damieng:
I think I might have dropped the ball on this one - do you want to rebase and pick it up?

Sjord:
I have more or less given up on this. I can't get conditional operators to work correctly. This PR is not sufficiently finished that it solves the precedence problems without introducing new problems. I'll close it.

PR #186, opened 2021-05-20
Support Unicode identifiers

Sjord:
According to documentation (https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names):

> Identifiers may contain Unicode letter characters, decimal digit characters, Unicode connecting characters, Unicode combining characters, or Unicode formatting characters.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1487,8 +1487,8 @@ module.exports = grammar({
       field('right', $._type)
     )),
 
-    // Unicode categories: L = Letter, Nd = Decimal_Number, Pc = Connector_Punctuation, Cf = Format, Mn = Nonspacing_Mark, Mc = Spacing_Mark, Me = Enclosing_Mark
-    _identifier_token: $ => token(seq(optional('@'), /[\p{L}_][\p{L}\p{Nd}\p{Pc}\p{Cf}\p{Mn}\p{Mc}\p{Me}]*/)),
+    // Unicode categories: L = Letter, Nl Letter_Number, = Nd = Decimal_Number, Pc = Connector_Punctuation, Cf = Format, Mn = Nonspacing_Mark, Mc = Spacing_Mark
+    _identifier_token: $ => token(seq(optional('@'), /[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Pc}\p{Cf}\p{Mn}\p{Mc}]*/)),
     identifier: $ => choice($._identifier_token, $._contextual_keywords),
 
     global: $ => 'global',
--- a/grammar.js
+++ b/grammar.js
@@ -1487,7 +1487,8 @@ module.exports = grammar({
       field('right', $._type)
     )),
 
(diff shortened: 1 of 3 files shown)


grammar.js
```diff
@@ -1487,7 +1487,8 @@ module.exports = grammar({
       field('right', $._type)
     )),
 
-    _identifier_token: $ => token(seq(optional('@'), /[a-zA-Zα-ωΑ-Ωµ_][a-zA-Zα-ωΑ-Ωµ_0-9]*/)), // identifier_token in Roslyn
+    // Unicode categories: L = Letter, Nd = Decimal_Number, Pc = Connector_Punctuation, Cf = Format, Mn = Nonspacing_Mark, Mc = Spacing_Mark, Me = Enclosing_Mark
+    _identifier_token: $ => token(seq(optional('@'), /[\p{L}_][\p{L}\p{Nd}\p{Pc}\p{Cf}\p{Mn}\p{Mc}\p{Me}]*/)),
```
damieng, line 1491: I'm not 100% sure the first letter can be any unicode letter. The page https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names that the first character has to be a _ or "a letter". It then goes on to say the next characters may contain "Unicode letter characters" which seems to be an intentional difference.

I tried `𐐣` in Visual Studio and it's okay after the first letter but not AS the first letter. I think we can either leave the first letter matcher as-is and adopt the rest or just go with what you have here and know we're being more relaxed than the real parser. I can roll with either so it's your call.

e.g.
```csharp
var 𐐣 = 1;
```
Gives an error in Visual Studio 2019 while:

```csharp
var _𐐣 = 1;
```

Does not.

maxbrunsfeld (approved):
I don't have a strong understanding of which exact unicode categories should be used for C#. But in general, this LGTM.

Sjord:
I am not sure why `var 𐐣 = 1` gives an error, while `var П = 1` works. From Unicode perspective, they are both capital letters. While looking into this, I came across the [roslyn code that handles Unicode in identifiers](https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/InternalUtilities/UnicodeCharacterUtilities.cs). It specifies Unicode categories which we can use.

I will look into what letters are acceptable starting letters a bit more.

Sjord:
It seems that at one time, C# specifically used Unicode 3.0. For example, [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#identifiers):

> For information on the Unicode character classes mentioned above, see The Unicode Standard, Version 3.0, section 4.5.

This would explain why `var 𐐣 = 1` doesn't work. 𐐣 was introduced in Unicode 3.1.

It seems that this dependency on Unicode 3.0 was removed from [more recent language specifications(https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/Unicode%20Version.md). Also, it would be pretty hard to implement this exactly in tree-sitter. So I think accepting all letters is acceptable.

damieng (approved):
Sounds good, thanks for the investigation as well.

Issue #189, opened 2021-05-25
Error on conditional compilation preprocessor directive


Sjord:
```
var a = new ArrayList()
{
    #if DEBUG
        Capacity = 3
    #else
        Capacity = 4
    #endif
};
```

Actual:

```
(compilation_unit [0, 8] - [7, 10]
  (global_statement [0, 8] - [7, 10]
    (local_declaration_statement [0, 8] - [7, 10]
      (variable_declaration [0, 8] - [7, 9]
        type: (implicit_type [0, 8] - [0, 11])
        (variable_declarator [0, 12] - [7, 9]
          (identifier [0, 12] - [0, 13])
          (equals_value_clause [0, 14] - [7, 9]
            (object_creation_expression [0, 16] - [7, 9]
              type: (identifier [0, 20] - [0, 29])
              arguments: (argument_list [0, 29] - [0, 31])
              initializer: (initializer_expression [1, 8] - [7, 9]
                (preprocessor_call [2, 12] - [2, 21]
                  (preprocessor_directive [2, 12] - [2, 15])
                  (identifier [2, 16] - [2, 21]))
                (assignment_expression [3, 16] - [5, 28]
                  left: (identifier [3, 16] - [3, 24])
                  (assignment_operator [3, 25] - [3, 26])
                  right: (assignment_expression [3, 27] - [5, 28]
                    left: (integer_literal [3, 27] - [3, 28])
                    (preprocessor_call [4, 12] - [4, 17]
                      (preprocessor_directive [4, 12] - [4, 17]))
                    (ERROR [5, 16] - [5, 24])
                    (assignment_operator [5, 25] - [5, 26])
                    right: (integer_literal [5, 27] - [5, 28])))
                (preprocessor_call [6, 12] - [6, 18]
                  (preprocessor_directive [6, 12] - [6, 18]))))))))))
foo.cs	0 ms	(ERROR [5, 16] - [5, 24])
```

Expected:
```
(compilation_unit [0, 8] - [8, 0]
  (global_statement [0, 8] - [7, 10]
    (local_declaration_statement [0, 8] - [7, 10]
      (variable_declaration [0, 8] - [7, 9]
        type: (implicit_type [0, 8] - [0, 11])
        (variable_declarator [0, 12] - [7, 9]
          (identifier [0, 12] - [0, 13])
          (equals_value_clause [0, 14] - [7, 9]
            (object_creation_expression [0, 16] - [7, 9]
[...]

PR #191, opened 2021-05-29
Improve preprocessing parsing

### LINKED ISSUE #16: Add structured_trivia
Right now pre-processor directives are handled in a simple block with some limitations:

1. Does not support all operations
2. Does not tokenize conditions, pragmas etc.
3. Allows whitespace before the directive

The Roslyn grammar.txt has a [structured_trivia](https://github.com/tree-sitter/tree-sitter-c-sharp/blob/master/grammar.txt#L1087) that would help address at least 1 & 2.  

This would replace `preprocessor_directive` in grammar.js

### LINKED ISSUE #188: Preprocessor directives don't interpret escape sequences
```cs
#line 200 "C:\test\new\user.cs"
```

Expected:

```
(compilation_unit [0, 0] - [0, 31]
  (preprocessor_call [0, 0] - [0, 31]
    (preprocessor_directive [0, 0] - [0, 5])
    (integer_literal [0, 6] - [0, 9])
    (string_literal [0, 10] - [0, 31])))
```

Actual:

```
(compilation_unit [0, 0] - [0, 31]
  (preprocessor_call [0, 0] - [0, 31]
    (preprocessor_directive [0, 0] - [0, 5])
    (integer_literal [0, 6] - [0, 9])
    (string_literal [0, 10] - [0, 31]
      (escape_sequence [0, 13] - [0, 15])
      (escape_sequence [0, 18] - [0, 20])
      (ERROR [0, 22] - [0, 30]
        (ERROR [0, 22] - [0, 24])))))
foo.cs	0 ms	(ERROR [0, 22] - [0, 30])
```

In C# strings, escape sequences such as `\t` for tab and `\n` for newlines can be used. However, no such escape sequences exist in preprocessor directives. For example, the following will report an error on `C:\test\new\program.cs` and not on `C:[tab]est[newline]ew\user.cs`.

```
#line 200 "C:\test\new\user.cs"
var a = ;
```

```
$ dotnet build
...
C:\test\new\user.cs(200,17): error CS1525: Invalid expression term ';' 
```

According to tree-sitter, a `preprocessor_call` can contain a `_literal`, which can be a `string_literal`, which can contain an `escape_sequence`.

damieng:
Fixes #188 as well as `#IF` containing anything more than a basic token.

It also provides a syntax tree for these statements partly modelled on the C one.

This does not attempt to solve #189 - I'll take a look at that separately (again based on the C one)

Fixes #16 too

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -33,7 +33,7 @@ module.exports = grammar({
   extras: $ => [
     $.comment,
     /[\s\u00A0\uFEFF\u3000]+/,
-    $.preprocessor_call
+    $._preprocessor_call
   ],
 
   supertypes: $ => [
@@ -1639,7 +1639,7 @@ module.exports = grammar({
     return_type: $ => choice($._type, $.void_keyword),
     void_keyword: $ => 'void',
 
-    preprocessor_call: $ => seq(
+    _preprocessor_call: $ => seq(
       $._preproc_directive_start,
       choice(
           $.nullable_directive,
--- a/grammar.js
+++ b/grammar.js
@@ -205,7 +205,7 @@ module.exports = grammar({
     ),
 
     attribute_argument: $ => seq(
-      optional(choice($.name_equals,$.name_colon)),
+      optional(choice($.name_equals, $.name_colon)),
       $._expression
     ),
 
@@ -1642,19 +1642,19 @@ module.exports = grammar({
     _preprocessor_call: $ => seq(
       $._preproc_directive_start,
       choice(
-          $.nullable_directive,
-          $.define_directive,
-          $.undef_directive,
-          $.if_directive,
-          $.else_directive,
-          $.elif_directive,
-          $.endif_directive,
-          $.region_directive,
-          $.endregion_directive,
-          $.error_directive,
-          $.warning_directive,
-          $.line_directive,
-          $.pragma_directive
+        $.nullable_directive,
+        $.define_directive,
+        $.undef_directive,
+        $.if_directive,
+        $.else_directive,
+        $.elif_directive,
+        $.endif_directive,
+        $.region_directive,
+        $.endregion_directive,
+        $.error_directive,
+        $.warning_directive,
+        $.line_directive,
+        $.pragma_directive
       ),
       $._preproc_directive_end
     ),
@@ -1662,9 +1662,9 @@ module.exports = grammar({
     _preproc_directive_start: $ => /#[ \t]*/,
 
     nullable_directive: $ => seq(
-        'nullable',
-        choice('disable', 'enable', 'restore'),
-        optional(choice('annotations', 'warnings'))
+      'nullable',
+      choice('disable', 'enable', 'restore'),
+      optional(choice('annotations', 'warnings'))
     ),
 
     // Preprocessor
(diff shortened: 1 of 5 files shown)


grammar.js
```diff
@@ -1640,16 +1640,101 @@ module.exports = grammar({
...
+    endif_directive: $ => 'endif',
+    region_directive: $ => seq('region', optional($.preproc_message)),
+    endregion_directive: $ => seq('endregion', optional($.preproc_message)),
+    error_directive: $ => seq('error', $.preproc_message),
+    warning_directive: $ => seq('warning', $.preproc_message),
+    line_directive: $ => seq('line',
+        choice(
+            'default',
+            'hidden',
+            seq($.preproc_integer_literal, optional($.preproc_string_literal))
```
Sjord, line 1686: This allows these two values without white space in between:

    #line 200"C:\test\new\user.cs"

This doesn't work in Roslyn. I don't think this is a big problem, however.

grammar.js
```diff
@@ -1640,16 +1640,101 @@ module.exports = grammar({
...
+        )
+    ),
+    pragma_directive: $ => seq('pragma',
+        choice(
+            seq('warning', choice('disable', 'restore'), commaSep($.preproc_warning_number)),
```
Sjord, line 1691: Apparently, besides configuring warnings on number, it's also possible to use identifiers. E.g. the following is valid:

    #pragma warning disable nullable

Sjord (commented):
* Apparently, there is something called ScriptCS which also has `#load` and `#r` preprocessor directives. I am not sure if we should support those.
* Roslyn supports a shebang `#!...` at the top of the file. Do we also want to support this?

grammar.js
```diff
@@ -1640,16 +1640,101 @@ module.exports = grammar({
...
+          $.error_directive,
+          $.warning_directive,
+          $.line_directive,
+          $.pragma_directive
+      ),
       $._preproc_directive_end
     ),
 
-    preprocessor_directive: $ => /#[ \t]*[a-z]\w*/,
+    _preproc_directive_start: $ => /#[ \t]*/,
```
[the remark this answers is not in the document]
damieng: I don't think we should support ScriptCS - it has a bunch of different grammar rules and with the top-level statements I think main CS will replace it now.

We should support that `#!` tho, I'll add there. Where did you spot that (in case I missed others)?

grammar.js
damieng: I'm okay with accepting we're a bit more relaxed there.

grammar.js
damieng: Good catch, should probably just make it accept any non-whitespace token.

grammar.js
```diff
@@ -1688,15 +1688,20 @@ module.exports = grammar({
     ),
     pragma_directive: $ => seq('pragma',
         choice(
-            seq('warning', choice('disable', 'restore'), commaSep($.preproc_warning_number)),
+            seq('warning',
+              choice('disable', 'restore'),
+              commaSep(
+                  choice(
+                    $.identifier,
+                    alias($.preproc_integer_literal, $.integer_literal),
```
Sjord, line 1696: This was `preproc_warning_number`, but you changed this to `preproc_integer_literal`. Was this intentional? Would it be possible to use warning identifiers such as "CS0219" here?

grammar.js
```diff
@@ -1688,15 +1688,20 @@ module.exports = grammar({
...
+            ))),
             seq('checksum', $.preproc_string_literal, $.preproc_string_literal, $.preproc_string_literal)
         )
     ),
 
-    preproc_message: $ => /.+/,
+    preproc_message: $ => /[^\n\r]*/,
```
Sjord, line 1702: Should this be `+` instead of `*`? Can a `preproc_message` be empty?

Sjord:
Should `_preproc_directive_start` check that it's the only thing on the line? I.e. this shouldn't be valid:

    var a = 3; #define A

grammar.js
damieng: Yes, good catch.

grammar.js
damieng: It allows identifier as well. I thought between identifier and integer we'd have good coverage but I should probably just put a regex in here.

damieng:
> Should _preproc_directive_start check that it's the only thing on the line? I.e. this shouldn't be valid:

It should, but I don't know how to do that with the way the grammar is setup right now. I'll tackle that separately once I've looked into the branching conditional logic handling ideas.

grammar.js
damieng: Hmm, I actually quite prefer the message numbers showing up as either an integer_literal or an identifier depending on whether they are purely numeric or start with letters so will stick with this unless there are objections.

PR #192, opened 2021-06-07
Rename the body of a record _record_base to _record_body

Sjord:
Having `_record_base` and `record_base` is confusing. Since it has an
underscore before it, it is not a public name, so this change shouldn't cause
much problems.


damieng (approved):
👍🏻

Issue #194, opened 2021-06-17
Comments allowed in string literals


Sjord:
```
string s = "\u0065/* \u0065 */\u0065";
```

Actual:
```
(compilation_unit [0, 0] - [0, 38]
  (global_statement [0, 0] - [0, 38]
    (local_declaration_statement [0, 0] - [0, 38]
      (variable_declaration [0, 0] - [0, 37]
        (predefined_type [0, 0] - [0, 6])
        (variable_declarator [0, 7] - [0, 37]
          (identifier [0, 7] - [0, 8])
          (equals_value_clause [0, 9] - [0, 37]
            (string_literal [0, 11] - [0, 37]
              (escape_sequence [0, 12] - [0, 18])
              (comment [0, 18] - [0, 30])
              (escape_sequence [0, 30] - [0, 36]))))))))
```

Expected:
```
(compilation_unit [0, 0] - [0, 38]
  (global_statement [0, 0] - [0, 38]
    (local_declaration_statement [0, 0] - [0, 38]
      (variable_declaration [0, 0] - [0, 37]
        (predefined_type [0, 0] - [0, 6])
        (variable_declarator [0, 7] - [0, 37]
          (identifier [0, 7] - [0, 8])
          (equals_value_clause [0, 9] - [0, 37]
            (string_literal [0, 11] - [0, 37]
              (escape_sequence [0, 12] - [0, 18])
              (escape_sequence [0, 21] - [0, 27])
              (escape_sequence [0, 30] - [0, 36]))))))))
```

Sjord:
Never mind, this was a mistake on my part.

PR #195, opened 2021-06-17
Add tests for block comments within strings

Sjord:
String literals should not interpret comments.

I thought this was a bug (see #194), but the grammer handled this correctly. I
already wrote the tests, and it is an interesting case, so might as well add
the tests.


Sjord:
See also #197.

damieng:
Looks great, thanks for increasing the coverage! :)

damieng (approved):

PR #196, opened 2021-06-17
Extract _string_literal_fragment and _interpolated_string_text_fragment

Sjord:
See https://github.com/returntocorp/semgrep/issues/3289

This results in a better name for string fragments (Str_lit_frag instead of
Imm_tok_pat_5a6fa79), and solves a problem where the node is not created in the
CST.


mjambon (approved):
This should make ocaml-tree-sitter happy. Thanks!

PR #206, opened 2021-09-07
Add support for file_scoped_namespace_declaration

damieng:
As per the C# 10 draft at https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/file-scoped-namespaces

Sticking with Roslyn naming and grammar.

Should probably also remove `global_statement` and replace it with just `statement` for consistency (in a subsequent PR if no objections)

### DIFF
--- a/corpus/source-file-structure.txt
+++ b/corpus/source-file-structure.txt
@@ -111,6 +111,24 @@ namespace A {
         (declaration_list)))))
 
 =====================================
+File scoped namespaces
+=====================================
+
+namespace A;
+
+class B {
+}
+
+---
+
+(compilation_unit
+  (file_scoped_namespace_declaration
+    (identifier)
+  (class_declaration
+    (identifier)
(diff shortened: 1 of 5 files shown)


damieng:
Any chance you can review this @initram ?

initram:
I think it looks good. I don't know why we have the global_statement, but it would be fine by be to replace that in a future PR.

initram (approved):

corpus/source-file-structure.txt
```diff
@@ -110,6 +110,24 @@ namespace A {
...
+
+class B {
+}
+
+---
+
+(compilation_unit
+  (file_scoped_namespace_declaration
+    (identifier)
+  (class_declaration
```
initram, line 127: just a nit, here it should be indented to show that the class is nested under the namespace declaration.

PR #207, opened 2021-09-09
Added support for record structs

initram:
See spec here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/record-structs

Syntax wise they are just records with an extra keyword, so I have only added one test to see that we can parse the basic version.


damieng:
Looks good to me, will just need a rebase and re-generate before merge.

Wonder if we should move `partial` out of the attributes in the future for better alignment with the official syntax. (This alignment is making keeping up with C# so much easier)

Issue #18, opened 2019-11-18
Add xml doc comment parsing


damieng:
There is a lot of XML specified in the grammar.txt but it is not clear if/how it ties in with the `///` commenting (there is no pattern for `///` listed)

amaanq:
This should be done via injections - this is similar to rust's doc comments where it can contain markdown, so with c# containing xml in its doc comments it should inject using the [xml parser](https://github.com/tree-sitter-grammars/tree-sitter-xml)

Injections are handled however downstream projects want to handle them, so this isn't really actionable here

PR #208, opened 2021-09-09
First attempt at XML Doc comment parsing

damieng:
First attempt at XML doc comment parsing referenced in #18 

I'm pretty happy with how this has turned out so far and it's quite in-line with the Roslyn way of parsing the basic types shown in the corpus. Specifically:

- `xml_element` is working nicely identifying start, end and gobbling up intermediate lines
- `xml_empty_element` works nicely and matches Roslyn here in naming/being a separate element
- `xml_cref_attribute` matching for identity highlighting crefs.

I need to go through and consider a few more things before this is ready to review:

1. `xml_text` appears multiple times in sequence where Roslyn emits a single instance. I suspect I need to make the `xml_text` rule greedier.
2. Should probably support `/**` multi-line XML doc comments as they're technically a thing... I have never seen actually used tho.
3. `xml_text_literal_token` probably needs a much better rule
4. The grammar fails to recover from unclosed xml tags but Roslyn does. This could be a poor editing experience. Need to figure out a way to recover without losing the `xml_element` multi-line stuff.
5. We can't know what type the identifier in the cname is as we have zero context. It could be a parameter name, a class name, etc. We just don't know so this won't be able to match Roslyn for lookup/syntax highlighting etc. without some kind of external symbol matching outside of this grammar.


initram:
I know this is a draft, but just keep in mind that technically it is only a doc-comment if there is exactly 3 slashes. If there are 4 or more slashes in a row, then it is just a normal single line comment. The same logic is there for multi-line comments, where 2 stars are doc-comments, 1 or 3+ is a normal comment.

But really cool to see you take a stab at this, I really had no clue how to approach this in the first place.

damieng:
Thanks, I'll bear that in mind! Yeah, I had taken at least 2 previous stabs at this and got nowhere. I had forgotten enough of the previous attempts to come at it from a fresh angle.

AnonymusRaccoon:
I'm kinda new to treesitter, but shouldn't the xml doc be considered as a xml injection? Treesitter won't be able to parse more than xml anyway, since it as no symbols awareness.

Expect from reducing duplication and having a simpler implementation, injecting the xml parser would allow tools that do things with xml to work. Like auto closing tags and so on.

damieng:
The XML doc comments aren't really a full XML system, we don't need to support namespaces etc. A simple subset like this PR would work fine - it's literally just 34 lines.

PR #212, opened 2021-09-20
Added support for extended property patterns

initram:
subpatterns now use expression_colon instead of name_colon. This alligns with how it is parsed in Roslyn, but resulted in me adding another conflict to the grammar.


damieng (approved):
Looks great!

PR #213, opened 2021-09-22
Test precedence between is and comparison operators

### LINKED ISSUE #161: Incorrect precedence between is and comparison operators
```
        var someObj = new object();
        var allowedValuesList = someObj is null
            ? default
            : new object();
```

Actual: `someObj is (null ? ... : ...)`
Expected: `(someObj is null) ? ... : ...`

Sjord:
Closes #161


damieng:
I don't think there's any need for the first line right? It's just the second we're testing. 

Code fragments in the corpus are okay being partial fragments that won't compile on their own.

PR #216, opened 2021-11-04
0.19.1

damieng:
Bump to get out all the great C# 10 work and many fixes for earlier versions.

@dcreager I think I can publish this to npm - do you need to do something for Rust/crate manually?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,15 @@
 [package]
 name = "tree-sitter-c-sharp"
 description = "C# grammar for the tree-sitter parsing library"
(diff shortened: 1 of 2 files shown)


phodal:
Hi @damieng @dcreager can you publish to Crate ? I build a project which depend on version 0.19, but it's never publish to Rust/crate after 0.16.1.

damieng:
Hopefully dcreager will be able to sort this out for you - I've not been involved in the Rust/crate side of things but would be happy to learn if he wants to pair on this part or send notes.

PR #220, opened 2022-03-21
Add an action to publish the grammar automatically

Luni-4:
This PR allows to publish automatically `tree-sitter-c-sharp` on crates.io when a new tag is set. To make it work, it is just necessary to define the `CARGO_REGISTRY_TOKEN` as secret.

It should be merged after #219 in my opinion

Thanks in advance for your review! :)

### DIFF
--- a/.github/workflows/publish_crate.yml
+++ b/.github/workflows/publish_crate.yml
@@ -1,0 +1,36 @@
+name: Publish on crates.io
+
+on:
+  push:
+    tags:
+      - v*
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
+
+jobs:
+  publish:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


damieng:
@dcreager I think you'd need to setup the secret - I don't have any access to the Rust/crates publishing only to the npm.

damieng:
Any chance you could pick this up @hendrikvanantwerpen ?

hendrikvanantwerpen:
I don't think I have Cargo access that allows me to do that. Perhaps @patrickt knows something about this as well.

Luni-4:
It should work now because of https://github.com/tree-sitter/tree-sitter-java/pull/98#issuecomment-1235934619. The token is public for the entire organization. Once merged, I guess you can publish through this action

Luni-4:
Thank you @damieng for the merge and the new release! It works! :tada:

damieng:
No, thank you for the script! Sorry things took a while to get in place.

PR #221, opened 2022-03-25
Tree sitter 0.20, Missing HIGHLIGHT_QUERY Rust binding

Tiggilyboo:
Not sure why the odd version logic in tree-sitter dependency, but when consuming this repository it still resolves to 0.19.*. Most of the other related language crates have seem to bumped to `0.20` explicitly, so I have done so also.

Additionally, upon consuming the crate, I noticed that the binding for `HIGHLIGHT_QUERY` was missing (also as in other crates), so I've also added that.

Let me know if I've missing something if there is a proper binding template to go from? New to all of this!

Cheers

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ include = [
 path = "bindings/rust/lib.rs"
 
 [dependencies]
-tree-sitter = ">= 0.19, < 0.21"
+tree-sitter = "0.20"
 
 [build-dependencies]
 cc = "1.0"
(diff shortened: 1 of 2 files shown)


damieng:
I'm just looking at putting out a new version and it seems that tree-sitter-javascript are specifying:

```
tree-sitter = ">= 0.19, < 0.21"
```

https://github.com/tree-sitter/tree-sitter-javascript/blob/master/Cargo.toml#L28

and some others too

https://github.com/tree-sitter/tree-sitter-java/blob/master/Cargo.toml#L28
https://github.com/tree-sitter/tree-sitter-ruby/blob/master/Cargo.toml#L28

It would seem prudent to change our line there to match that rather than pinning to 0.20 before release.

PR #224, opened 2022-04-30
C# 11 features

### LINKED ISSUE #223: C# 11.0 feature support
Issue to track the various parts of C# 11 support - currently in preview so subject to change before then.

Some C# 11 syntax works already as we are not as strict as Roslyn, specifically:

- [x] Generic attributes
- [x] Static abstract interface methods
- [x] Newlines in interpolated strings

Some remaining support is required for:

- [x] List patterns
- [x] Slice patterns
- [ ] Raw string literals #222

damieng:
Fixes #223 partially by:

- Adding [List and Slice Patterns](https://github.com/dotnet/csharplang/blob/main/proposals/list-patterns.md)
- Adding [UTF-8 Literals](https://github.com/dotnet/csharplang/blob/main/proposals/utf8-string-literals.md)

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -1877,9 +1877,9 @@ var x = c is < '0' or >= 'A' and <= 'Z';
           (equals_value_clause
             (is_pattern_expression
               (identifier)
-              (binary_pattern
+              (or_pattern
                 (relational_pattern (character_literal))
-                (binary_pattern
+                (and_pattern
                   (relational_pattern (character_literal))
                   (relational_pattern (character_literal)))))))))))
 
@@ -2019,7 +2019,7 @@ var c = o is int; //is_expression with type
           (equals_value_clause
             (is_pattern_expression
               (identifier)
-              (binary_pattern
+              (or_pattern
                 (type_pattern (predefined_type))
                 (type_pattern (predefined_type))))))))) (comment)
   (global_statement
@@ -2033,6 +2033,31 @@ var c = o is int; //is_expression with type
                 (identifier)
                 (predefined_type))))))) (comment))
 
+=====================================
+List patterns
+=====================================
+
+var a = p is [1,2,x,] and [] or [2,..];
+
+---
+
+(compilation_unit
+  (global_statement
+    (local_declaration_statement
+      (variable_declaration (implicit_type)
+        (variable_declarator (identifier)
+          (equals_value_clause
+            (is_pattern_expression (identifier)
+              (or_pattern
+                (and_pattern
+                  (list_pattern
+                    (constant_pattern (integer_literal))
+                    (constant_pattern (integer_literal))
+                    (constant_pattern (identifier)))
+                  (list_pattern))
+                  (list_pattern
+                    (constant_pattern (integer_literal))
+                    (slice_pattern))))))))))
 
 =====================================
 Conditional expression with member accesses
--- a/corpus/literals.txt
+++ b/corpus/literals.txt
@@ -314,6 +314,50 @@ class A {
           (identifier)
           (equals_value_clause (string_literal (escape_sequence)))))))))
 
+=======================================
+utf-8 string literals
+=======================================
+
+var a = "lower"u8;
+var b = "upper"U8;
+var c = @"lower"u8;
+var d = @"upper"U8;
+---
+
+ (compilation_unit
+    (global_statement
+      (local_declaration_statement
+        (variable_declaration
+          (implicit_type)
+          (variable_declarator
+            (identifier)
+            (equals_value_clause
+              (string_literal))))))
+    (global_statement
+      (local_declaration_statement
+        (variable_declaration
+          (implicit_type)
+          (variable_declarator
+            (identifier)
+            (equals_value_clause
+              (string_literal))))))
+    (global_statement
+      (local_declaration_statement
+        (variable_declaration
+          (implicit_type)
+          (variable_declarator
+            (identifier)
+            (equals_value_clause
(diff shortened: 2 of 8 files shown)


corpus/expressions.txt
```diff
@@ -1877,9 +1877,9 @@ var x = c is < '0' or >= 'A' and <= 'Z';
           (equals_value_clause
             (is_pattern_expression
               (identifier)
-              (binary_pattern
+              (or_pattern
                 (relational_pattern (character_literal))
-                (binary_pattern
+                (and_pattern
```
hendrikvanantwerpen, line 1882: What is the rationale for removing `binary_pattern`? It looks like the change doesn't change which programs are accepted, and changing it potentially breaks any existing consumers of the syntax tree. I think this change should be reverted.

corpus/literals.txt
```diff
@@ -314,6 +314,50 @@ class A {
...
           (equals_value_clause (string_literal (escape_sequence)))))))))
 
+=======================================
+utf-8 string literals
+=======================================
+
+var a = "lower"u8;
+var b = "upper"U8;
+var c = @"lower"u8;
+var d = @"upper"U8;
```
hendrikvanantwerpen, line 324: A reference would be helpful here. I couldn't find this syntax in previews of the upcoming C# 11 features.

grammar.js
```diff
@@ -922,12 +923,22 @@ module.exports = grammar({
...
+      $.and_pattern,
+      $.list_pattern,
       $.type_pattern
     ),
 
     type_pattern: $ => $._type,
 
+    list_pattern: $ => seq(
+      '[',
+      optional(seq(commaSep(choice($._pattern, $.slice_pattern)), optional(','))),
```
hendrikvanantwerpen, line 936: Should this be `commaSep1`? Otherwise, a single comma without preceding elements is allowed, I think.

hendrikvanantwerpen (changes requested):
First of all, thanks for contributing to Tree-sitter!

Since C# 11 hasn't been released yet, I think it is better to wait until the official release, so it's clear what actually got in and in what form.

But, since you marked this as ready for review, I'll leave a couple of questions already. The general one is to add some links to official docs describing the new syntax in the PR. This helps a lot to review it, since not all the reviewers are always up to date with the latest language changes.

A few more comments below.

corpus/literals.txt
damieng: https://github.com/dotnet/csharplang/blob/main/proposals/utf8-string-literals.md is where the details can be found. Most of the proposals for C# start in that repo and you can see it was approved in [January 2022](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-01-26.md#suffix-case-sensitivity) and has been available in Visual Studio 2022 Beta for several months.

corpus/expressions.txt
damieng: Up until this point there hasn't been a specific goal on tree-sitter-c-sharp to retain every published named rule.

If that's a goal across tree-sitter i is mostly doable but introduces extra complexity in trying to keep up with new C# grammar rules where right now we try and stick close to the Roslyn grammar (as tricky as that is to piece together given that the official spec is always many versions behind the real language).

hendrikvanantwerpen: Thanks for clarifying! I think these things are somewhat underspecified in the tree-sitter ecosystem atm :) Alignment with some reference is definitely valuable, so if this achieves that, I think that's a valid reason to go through with it.

corpus/literals.txt
hendrikvanantwerpen: :+1: My search was definitely too shallow to uncover that :)

grammar.js
damieng: Good catch, I must have thought `[,]` was legal syntax but it is not so I can switch that out.

corpus/literals.txt
damieng: Yeah keeping up with the C# syntax is not a nice experience :( Blog posts, proposals, stuff in and out of VS and Roslyn... I think these two are solid based on the discussion/lack of controversy but could hold off another month.

corpus/expressions.txt
damieng: Okay, dug a bit more into this (the PR was in draft for a while so I'd forgotten most of it).

The grammar such as it is documented has kept the `binary_pattern` for or and and *but* I was unable to get them to parse correctly as I needed to deal with precedence so split them so each could have their own.

In the VS Syntax Analyzer (primary way I figure out what Rolsyn is *actually* doing) I see that `OrPattern` and `AndPattern` are actually separately specified.

e.g.

```csharp
var a = p is [1, 2, x,] and [] or [2, ..];
```

Corresponds to 
image

hendrikvanantwerpen: That makes sense. And thanks for the extensive explanation ;)

corpus/literals.txt
hendrikvanantwerpen: I don't know what usually happened? You said it is already available in VS Beta. If people can write it, that could be an argument to already merge it?

grammar.js
damieng: Fixed in latest commit.

hendrikvanantwerpen (approved):
Looks good!

I see issue #223 got automatically linked to the PR. Would merging this close that, or is GH smart enough to consider the "partially" that comes after?

damieng:
GH will probably close the linked issue but I think that's okay, I already span out another ticket for the remaining complex C# 11 feature.

PR #229, opened 2022-08-22
Added missing record struct highlight

votagus:


damieng:
Hi, thanks for your pull request and fix.

Can you add a small snippet to demonstrate this parsing scenario to the most appropriate .txt file in the corpus folder please - they serve as tests for this project. (There are a lot of examples)

votagus:
Hi, I found this in the file records.txt
```
=====================================
Basic record struct declaration
=====================================

record struct F {
  int Age { get; init; }
}

---

(compilation_unit
  (record_struct_declaration
    (identifier)
    (declaration_list
      (property_declaration
        (predefined_type)
        (identifier)
        (accessor_list (accessor_declaration) (accessor_declaration))))))
```
If that does not already cover the scenario or is not what you asked for I think we will have to wait since I barely started with tree-sitter and don't know much yet.

damieng:
Sorry, just realized you're editing the highlights file not the language grammar - no need for corpus text there so this is probably good to go.

Issue #230, opened 2022-09-01
Parse error when record ends with semicolon


Sjord:
The second example on the [Records page in the documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) has a record ending in a semicolon:

```
public record Person
{
    public string FirstName { get; init; } = default!;
    public string LastName { get; init; } = default!;
};
```

This causes an error when parsed with tree-sitter:

```
(compilation_unit [0, 3] - [5, 0]
  (record_declaration [0, 3] - [4, 1]
    ...)
  (ERROR [4, 1] - [4, 2]))
```

Issue #225, opened 2022-05-01
Parse error when attribute is added on lambda


romamihalich:
Valid C# code:
```C#
var hi = [Foo] () => Console.WriteLine("hi");

public class FooAttribute : System.Attribute {}
```
Executing command below results in error
```Console
$ tree-sitter parse Program.cs
```
Output:
```
(compilation_unit [0, 3] - [3, 0]
  (global_statement [0, 3] - [0, 48]
    (local_declaration_statement [0, 3] - [0, 48]
      (variable_declaration [0, 3] - [0, 47]
        type: (implicit_type [0, 3] - [0, 6])
        (variable_declarator [0, 7] - [0, 47]
          (identifier [0, 7] - [0, 9])
          (equals_value_clause [0, 10] - [0, 47]
            (invocation_expression [0, 12] - [0, 47]
              function: (member_access_expression [0, 12] - [0, 41]
                expression: (invocation_expression [0, 12] - [0, 20]
                  function: (element_binding_expression [0, 12] - [0, 17]
                    (bracketed_argument_list [0, 12] - [0, 17]
                      (argument [0, 13] - [0, 16]
                        (identifier [0, 13] - [0, 16]))))
                  arguments: (argument_list [0, 18] - [0, 20]))
                (ERROR [0, 21] - [0, 31])
                name: (identifier [0, 32] - [0, 41]))
              arguments: (argument_list [0, 41] - [0, 47]
                (argument [0, 42] - [0, 46]
                  (string_literal [0, 42] - [0, 46])))))))))
  (class_declaration [2, 0] - [2, 47]
    (modifier [2, 0] - [2, 6])
    name: (identifier [2, 13] - [2, 25])
    bases: (base_list [2, 26] - [2, 44]
      (qualified_name [2, 28] - [2, 44]
        (identifier [2, 28] - [2, 34])
        (identifier [2, 35] - [2, 44])))
    body: (declaration_list [2, 45] - [2, 47])))
Program.cs	0 ms	(ERROR [0, 21] - [0, 31])
```

damieng:
Looks like we missed something on the C# 10 support, I'll get it added.

PR #231, opened 2022-09-01
Fix Lambda attributes and record trailing semicolons

damieng:
Fixes:

- #225 
- #230


damieng:
This one is a non-controversial bug fix so I'm going to sail it through.

PR #238, opened 2022-09-22
Added support for file scoped classes

### LINKED ISSUE #236: C# 11 file scoped types
They seem to still be adding features to C# 11.

Docs can be found here:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/file

initram:
Fixes #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Comprehensive supports C# 1 through 10.0 with the following exceptions:
 - [x] List patterns
 - [x] Slice pattern
 - [x] Required members
+- [x] File scoped classes
 - [ ] Raw string literals
 
 ### References
--- a/corpus/classes.txt
+++ b/corpus/classes.txt
@@ -560,3 +560,18 @@ class Ωµ {
         parameters: (parameter_list)
         body: (block
           (return_statement (null_literal)))))))
+
+=================================
+File scoped class
+=================================
+
+file class A {}
+
+---
+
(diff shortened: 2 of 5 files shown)


damieng:
It looks like this is interfering with variables named "file" as we re-use the `modifier` rules in a lot of places.

You might be able to do the `prec(1, ` trick we did for `ref` - failing that trying modifying `repeat($.modifier),` on `class_declaration` etc. to be `repeat(choice($.modifier, 'file'))` on the relevant possibilities (the C# guide doesn't seem comprehensive on applicability)

damieng:
Tried both of those and they involve other types of conflict. Another option might be to split our modifier types up.

initram:
I don't know if it will make sense to create a separate identifier for types, that is different from the one for variables and members. I see that in the latest version of the compiler they have made changes that disallow types to be named "file", "scoped", or "required".

Maybe that would help solve some of the issues we have with parsing these new keywords. But it would be a rather large change to the output returned, unless there is some way of doing it transparently.

damieng:
I think we could split the parsing up into two different types and then alias them back into the original one unless I'm missing something.

The question I think is whether this would solve the problems or cause more - in the past it's been hard to predict either way.

tamasvajk:
I've rebased the PR and pushed an extra commit to it in https://github.com/tree-sitter/tree-sitter-c-sharp/pull/257.

PR #240, opened 2022-11-04
Raw string literals support

### LINKED ISSUE #222: c# 11 raw string literals
per [docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated), these are valid and invalid cases starting with c# 11
```c#
// valid
x = $"""The point "{X}, {Y}" is {Math.Sqrt(X * X + Y * Y)} from the origin""";
x = $"""The point ""{X}, {Y}"" is {Math.Sqrt(X * X + Y * Y)} from the origin""";
x = $""""The point """{X}, {Y}""" is {Math.Sqrt(X * X + Y * Y)} from the origin"""";
x = $$"""The point {{{X}}, {{Y}}} is {{Math.Sqrt(X * X + Y * Y)}} from the origin""";

// invalid
x = $"The point "{X}, {Y}" is {Math.Sqrt(X * X + Y * Y)} from the origin";
x = $""The point ""{X}, {Y}"" is {Math.Sqrt(X * X + Y * Y)} from the origin"";
x = $"""The point """{X}, {Y}""" is {Math.Sqrt(X * X + Y * Y)} from the origin""";
x = $$"""The point {{X}, {{Y}}} is {{Math.Sqrt(X * X + Y * Y)}} from the origin""";
```

also take a look at 'what's new in c# 11' series: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#raw-string-literals

tree-sitter doesn't handle valid cases.

damieng:
Fixes #222 by including support for:

- Raw string literals `"""hello ""bob"" """`
- Raw string literals utf-8 `"""hello ""bob"" """u8;`
- Interpolated raw string literals `$"""hello ""{name}"""."";`

Supporting spanning multiple lines. 

### Does not support an unlimited number of opening and closing double quotes

I have no idea how we can do this - the number of closing quotes must match the number of opening quotes which is not a pattern I've seen before in tree-sitter grammars (I don't think it's supported?). Anything inside this with *less* quotes than the opening number of quotes is valid literal text. - any thoughts @maxbrunsfeld  ?

e.g. the following is valid

```csharp
var x = $""""""""""""""""""The point """{X}, {Y}"""""" 
 is {Math.Sqrt(X * X + Y * Y)} from the origin$"""""""""""""""""";
```

and would emit

```The point """1, 2"""""" is 2.23606797749979 from the origin```

### Quotes inside literal emit additional `interpolated_raw_string_text`

Would be nice to eliminate this (somehow) but I can live with it otherwise.

### DIFF
--- a/corpus/literals.txt
+++ b/corpus/literals.txt
@@ -768,3 +768,268 @@ var x = $@"/* {world} */";
               (interpolation
                 (identifier))
               (interpolated_verbatim_string_text))))))))
+
+==================================================
+Raw string literal (basic)
+==================================================
+
+var x = """Hello"""; 
+
+---
+
+(compilation_unit
+  (global_statement
+    (local_declaration_statement
+      (variable_declaration
+        (implicit_type)
+        (variable_declarator
+          (identifier)
+          (equals_value_clause
+            (raw_string_literal)))))))
+
+==================================================
+Raw string literal (multi-line)
+==================================================
+
+var x = """
+  Hello
+  """; 
+
+---
+
+(compilation_unit
+  (global_statement
+    (local_declaration_statement
+      (variable_declaration
+        (implicit_type)
+        (variable_declarator
+          (identifier)
+          (equals_value_clause
+            (raw_string_literal)))))))
+
+==================================================
+Raw string literal (embedded one double quotes)
+==================================================
+
+var x = """
+  Hello "Damien"
+  """; 
+
+---
+
+(compilation_unit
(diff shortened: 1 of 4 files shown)


damieng:
Unfortunately no feedback so going to merge to align with the .NET 7 release with the known limitations for now.

alexpovel:
Hi! Thanks for this implementation. Is there a release planned to get this out the door?

PR #244, opened 2022-12-15
Add support for `static` modifier on anonymous method expressions

tamasvajk:


grammar.js
```diff
@@ -1093,7 +1093,7 @@ module.exports = grammar({
     ),
 
     anonymous_method_expression: $ => seq(
-      optional('async'),
+      optional(alias(choice('async', 'static', seq('async', 'static'), seq('static', 'async')), $.modifier)),
```
tamasvajk, line 1096: Why doesn't `repeat($.modifier)` work here?

damieng: We've had a lot of problems with modifiers - they are so broad that in many places it causes conflicts while allowing invalid syntax for specific scenarios (which I think Roslyn flags in a second-stage)

damieng (approved):

damieng:
Thanks for the contribution!

Issue #247, opened 2022-12-16
Improve type parsing


tamasvajk:
Type parsing doesn't seem to handle all the possible cases that can come up in C#.

1. One particular case which is missing is `void*`. `void` is handled separately from all other types, and is only introduced as a return type, but that means that pointer types don't cover `void`.  `void*`, `void**`, ... can show up anywhere a type can show up, not just as a return type. I thing it would be the most elegant if `void` could be added to `predefined_type` as it is done in [Roslyn](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4#L420).

2. Similarly, it feels like shortcuts were made for ref types too. https://github.com/tree-sitter/tree-sitter-c-sharp/pull/246 allows `ref (readonly)` in delegate declarations, but that fix seems more like a workaround than a proper fix. For example in the below C# example, I think `ref` and the second `readonly` should actually belong to the `type` and not the `field_declaration`.
```C#
ref struct X {
    public X(){}
    readonly ref readonly int x = 5;
}
```
 produces
```
...
      field_declaration [2, 4] - [2, 36]
        modifier [2, 4] - [2, 12]
        modifier [2, 13] - [2, 16]
        modifier [2, 17] - [2, 25]
        variable_declaration [2, 26] - [2, 35]
          type: predefined_type [2, 26] - [2, 29]
          variable_declarator [2, 30] - [2, 35]
...
```

damieng:
Yeah I started adding ref_type - we have an unused stub for it - but adding it to the parser just exploded a whole bunch of conflicts so it was set aside as it wasn't a common pattern. We definitely have some precedent issues that I think are all tied together.

tamasvajk:
The two problems mentioned above have been solved.

PR #251, opened 2022-12-20
Add `ref`/`ref readonly` types to the type system

tamasvajk:
This PR partially covers https://github.com/tree-sitter/tree-sitter-c-sharp/issues/247.

Previously `public ref readonly int M()` was parsed with 3 modifiers on the method declaration, but this didn't match the semantics of the modifiers. Ref types are introduced in this PR, so the above method declaration parses with a single `public` modifier, and its return type is a `ref_type`.

I [tried solving](https://github.com/tree-sitter/tree-sitter-c-sharp/compare/master...tamasvajk:tree-sitter-c-sharp:fix/ref-type?expand=1) this issue with modifying the precedence of parameters vs types (and arguments vs ref_expressions), but that would require changing many other rule precedences. (I gave up on that approach). Instead, this PR removes `ref` from the general modifier list, and adjusts minimally the precedences.

### DIFF
--- a/corpus/type-fields.txt
+++ b/corpus/type-fields.txt
@@ -123,7 +123,7 @@ Function pointer type
 class A {
   // Function pointer equivalent without calling convention
   delegate*<string, int> a;
-  delegate*<delegate*<in string, int>, delegate*<string, ref readonly int>> b;
+  delegate*<delegate*<in string, int>, delegate*<ref string, ref readonly int>> b;
 
   // Function pointer equivalent with calling convention
   delegate* managed<string, int> c;
@@ -141,19 +141,19 @@ class A {
         (variable_declaration
           (function_pointer_type
             (function_pointer_parameter (predefined_type))
-            (function_pointer_parameter (predefined_type)))
+            (function_pointer_return_type (predefined_type)))
           (variable_declarator (identifier))))
       (field_declaration
         (variable_declaration
           (function_pointer_type
             (function_pointer_parameter
               (function_pointer_type
-                (function_pointer_parameter (predefined_type))
-                (function_pointer_parameter (predefined_type))))
-            (function_pointer_parameter
+                (function_pointer_parameter (parameter_modifier) (predefined_type))
+                (function_pointer_return_type (predefined_type))))
+            (function_pointer_return_type
               (function_pointer_type
-                (function_pointer_parameter (predefined_type))
-                (function_pointer_parameter (predefined_type)))))
+                (function_pointer_parameter (parameter_modifier) (predefined_type))
+                (function_pointer_return_type (ref_type (predefined_type))))))
           (variable_declarator (identifier))))
       (comment)
       (field_declaration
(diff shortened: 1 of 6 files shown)


corpus/type-fields.txt
```diff
@@ -141,27 +141,27 @@ class A {
...
       (field_declaration
         (variable_declaration
           (function_pointer_type
             (function_pointer_parameter
               (function_pointer_type
-                (function_pointer_parameter (predefined_type))
-                (function_pointer_parameter (predefined_type))))
-            (function_pointer_parameter
+                (function_pointer_parameter (parameter_modifier) (predefined_type))
+                (function_pointer_return_type (predefined_type))))
```
damieng, line 152: Both types appear as FunctionPointParameter in Roslyn, it would be nice to align a little closer here, some options might be:

1. Alias `function_pointer_return_type` so it appears as a `function_pointer_parameter` 
2. Rename `function_pointer_return_type` to `function_pointer_parameter_return_type` (or something that keeps parameter in there)

Thoughts?

grammar.js
```diff
@@ -244,7 +253,7 @@ module.exports = grammar({
       ';'
     )),
 
-    modifier: $ => prec.right(choice(
+    _modifier_no_ref: $ => prec.right(choice(
```
damieng, line 256: Maybe we just call this modifier and remove the other two to avoid the conflict rules?

grammar.js
```diff
@@ -572,7 +588,7 @@ module.exports = grammar({
 
     struct_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(alias($.modifier_with_ref, $.modifier)),
```
damieng, line 591: I think we can just go with `repeat($.modifier)` here followed by an `optional('ref')` before the `struct` as it looks like it has to be the last modifier before struct anyway.

damieng (changes requested):

corpus/type-fields.txt
tamasvajk: Option 1 (aliasing) sounds good. I did it in https://github.com/tree-sitter/tree-sitter-c-sharp/pull/251/commits/72b525e30ee8665cae8b92b2585ab9c727a85a28.

grammar.js
tamasvajk: Thanks. Sounds good. I've modified this and the above in https://github.com/tree-sitter/tree-sitter-c-sharp/pull/251/commits/08fecfd9b2ccc9743f70968136804626b5e05bbc.

damieng (approved):
Looks great!

tamasvajk:
@damieng Do you usually do any larger scale testing for changes that might have a broad impact?

damieng:
The CI does a full parse against NUnit, JSON.NET and Orchard albeit older versions that don't exercise a lot of the newer paths. 

If there is a regression in the ability to parse those files it will cause a regression but we don't do anything with the parse tree itself.

The relevant scripts are in the scripts folder (you can run them yourself too) along with the known failures files - all three are combinations of `#if` where you would have to parse the whole block multiple times following each possible combination to adequately check which tree-sitter didn't support.

PR #253, opened 2022-12-20
WIP: Add optional lambda return type

tamasvajk:


corpus/expressions.txt
```diff
@@ -812,8 +826,8 @@ MyIntDelegate a = (ref int i) => i + 1;
             (lambda_expression
               (parameter_list
                 (parameter
-                  (parameter_modifier)
-                  (predefined_type)
+                  (ref_type
+                    (predefined_type))
```
tamasvajk, line 830: I don't get why this changed. Seemingly it's at the same location in the parse tree, it's a parameter inside a lambda declaration. Why did the parsing order change?

damieng: I think it's because previously it was seeing it as a parameter but now you've added the optional type there and then previously added ref types it matches the ref_type before it matches the parameter.

damieng: Hmm, although the parameter part of the tree is still there so that's confusing.

tamasvajk:
Closing this as https://github.com/tamasvajk/tree-sitter-c-sharp/pull/2 seems a better approach.

PR #256, opened 2022-12-21
Remove all precedence from type system

tamasvajk:
This PR makes some fundamental changes to the type system. All precedence specifications are removed. The type system is made tighter, so `ref ref int` is not valid any longer, similarly to `int??`, which was already not allowed. Furthermore the parameter in `void M(ref ref int i){}` is also not parsing correctly any longer. Additionally, pointer types are removed from nullable base types, so `int*?` is not allowed any longer.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -77,10 +77,13 @@ module.exports = grammar({
     [$._type, $.stack_alloc_array_creation_expression],
     [$._type, $._nullable_base_type],
     [$._type, $._array_base_type],
+    [$._type, $._pointer_base_type],
     [$._type, $._nullable_base_type, $.array_creation_expression],
     [$._type, $._array_base_type, $.array_creation_expression],
+    [$._type, $._pointer_base_type, $.array_creation_expression],
     [$._nullable_base_type, $.stack_alloc_array_creation_expression],
     [$._array_base_type, $.stack_alloc_array_creation_expression],
+    [$._pointer_base_type, $.stack_alloc_array_creation_expression],
 
     [$.parameter, $.this_expression],
     [$.parameter, $._simple_name],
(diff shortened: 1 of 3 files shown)


damieng:
Sorry, I cleaned up some of the unused conflicts earlier - I can revert that for now if that helps you get unblocked.

Do you think these changes might help us move forward with addressing some of the other open precedence bugs?

tamasvajk:
> Sorry, I cleaned up some of the unused conflicts earlier - I can revert that for now if that helps you get unblocked.

No problem. I already rebased.

> Do you think these changes might help us move forward with addressing some of the other open precedence bugs?

Yes, that's my hope. It will definitely help in introducing `scoped` types, and hopefully I can also remove the precedence on `modifier`, which might help with introducing `file` modifier.

tamasvajk:
Ohh, for the `file` modifier, I found an easier solution: https://github.com/tree-sitter/tree-sitter-c-sharp/pull/257

tamasvajk:
Looks like this change helps with the optional return type of lambdas: https://github.com/tamasvajk/tree-sitter-c-sharp/pull/2.

tamasvajk:
I've done some more tests on this by adding the below to `fetch-examples` and checking the number of errors. Both before and after these changes, the count is 177. So I don't think this PR introduces any (major) regression.
```
checkout examples/roslyn dotnet/roslyn ac92012c0e64c8114ecf0024c2bad2f63dcc7c6b
```

damieng (approved):

damieng:
I wonder if we should add Roslyn to the CI test suite.

tamasvajk:
> I wonder if we should add Roslyn to the CI test suite.

That would be great, but we would need some improvement to `parse-examples`, because if a known failure is fixed, we are not forced to remove the entry from `known-failures.txt`.

PR #259, opened 2022-12-22
Add optional lambda return type

tamasvajk:


tamasvajk:
Issue count in Roslyn goes down to 175 (from 177)

damieng (approved):

PR #261, opened 2022-12-22
Add support for `scoped` types

tamasvajk:


grammar.js
```diff
@@ -753,6 +757,7 @@ module.exports = grammar({
     ),
 
     function_pointer_parameter: $ => seq(
+      alias(optional('scoped'), $.parameter_modifier),
```
tamasvajk, line 760: I just realized `scoped` is not a valid modifier on function pointer.

damieng (approved):

Issue #243, opened 2022-12-06
Generic type in declaration parsed as binary expression


Sjord:
```
IEnumerable<Clz> a = new Clz[1];
```

Expected:

```
(compilation_unit [0, 0] - [0, 32]
  (global_statement [0, 0] - [0, 32]
    (local_declaration_statement [0, 0] - [0, 32]
      (variable_declaration [0, 0] - [0, 31]
        type: (generic_name [0, 0] - [0, 16]
          (identifier [0, 0] - [0, 11])
          (type_argument_list [0, 11] - [0, 16]
            (predefined_type [0, 12] - [0, 15])))
        (variable_declarator [0, 17] - [0, 31]
          (identifier [0, 17] - [0, 18])
          (equals_value_clause [0, 19] - [0, 31]
            (array_creation_expression [0, 21] - [0, 31]
              (array_type [0, 25] - [0, 31]
                type: (predefined_type [0, 25] - [0, 28])
                rank: (array_rank_specifier [0, 28] - [0, 31]
                  (integer_literal [0, 29] - [0, 30]))))))))))
```

Actual:

```
(compilation_unit [0, 0] - [0, 32]
  (global_statement [0, 0] - [0, 32]
    (expression_statement [0, 0] - [0, 32]
      (assignment_expression [0, 0] - [0, 31]
        left: (binary_expression [0, 0] - [0, 18]
          left: (binary_expression [0, 0] - [0, 15]
            left: (identifier [0, 0] - [0, 11])
            right: (identifier [0, 12] - [0, 15]))
          right: (identifier [0, 17] - [0, 18]))
        (assignment_operator [0, 19] - [0, 20])
        right: (array_creation_expression [0, 21] - [0, 31]
          (array_type [0, 25] - [0, 31]
            type: (identifier [0, 25] - [0, 28])
            rank: (array_rank_specifier [0, 28] - [0, 31]
              (integer_literal [0, 29] - [0, 30]))))))))
```

PR #269, opened 2023-01-10
Split expression rule to improve parsing conflicting input

tamasvajk:
This PR splits expressions into two main categories based on whether they can or cannot be on the left side of assignments. Additionally the underlying expression types of expression statements are also limited.

These make the parser more restrictive than before. The following are not accepted any longer:
```C#
a < b;
m() = 0;
```

As a result, the PR fixes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/243. Initially `A<B> C = null;` was parsed as `((A < B) > C) = null;`. By not allowing comparison to be on the left side of assignments, the parse tree changes to `A < (B > (C = null));`, and by not allowing comparisons to be expression statements, the parse tree becomes a generic variable declaration `A<B> C = null;`. This is added as [a test case](https://github.com/tree-sitter/tree-sitter-c-sharp/pull/269/files#diff-9fc0f8e85b9fec5af062d8d60144a45bf395006ac261b5388fccc77f2f0271a3R328-R348).

There are some slightly adjusted test cases in the PR, those were using invalid expression statements.

There's at least one drawback to this approach:
```
var x = 
#if true
  0;
#else 
  1;
#endif
```
is not parsed without an error any longer. I think previously it was basically equivalent to
```
var x =
  0;
  1;
```
where `1;` was an expression statement. But that's not a valid expression statement any longer. We're basically trading an incorrectly parsed input to a failed parsing.

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -325,12 +325,114 @@ a >>>= a;
         (identifier)))))
 
 ================================================================================
+Assignment LValue types
+================================================================================
+
+a = 1;
+a.b = 1;
+a[b] = 1;
+(a, b) = (1, 2);
+(var a, b) = (1, 2);
+var x = new A
+{
+  a = 1,
+  [b] = 1
+};
+(a) = 1;
+
+--------------------------------------------------------------------------------
+
+(compilation_unit
+  (global_statement
+    (expression_statement
+      (assignment_expression
+        (identifier)
+        (assignment_operator)
+        (integer_literal))))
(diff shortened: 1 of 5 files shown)


damieng (approved):
This looks great and I think it's going to really help address all sorts of conflict cases thank-you so much for figuring this out.

I wouldn't worry about the `#if` pre-processor situation - our current parsing only really handled those cases by mistake more than anything else and we need a better all-round solution there.

PR #270, opened 2023-01-10
Remove unnecessary conflict rules

damieng:
Just removes some unnecessary conflict rules we don't need any more and also sets the min version of tree-sitter-cli so we don't regress back from languageversion 14.


damieng:
Will re-attempt this after all the various PRs in progress.

PR #272, opened 2023-01-11
Fix `_lvalue_expression` to include pointer indirection

tamasvajk:
I missed this case in https://github.com/tree-sitter/tree-sitter-c-sharp/pull/269.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -112,6 +112,8 @@ module.exports = grammar({
     [$.constant_pattern, $._lvalue_expression],
     [$.constant_pattern, $._expression_statement_expression],
 
+    [$.assignment_expression, $._expression],
+
   ],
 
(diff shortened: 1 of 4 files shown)


grammar.js
```diff
@@ -1254,7 +1256,7 @@ module.exports = grammar({
       ))
     )),
 
-    element_access_expression: $ => prec.right(PREC.UNARY, seq(
+    element_access_expression: $ => prec.right(PREC.POSTFIX, seq(
```
tamasvajk, line 1259: This is fixed to not have conflict in `*p[i]`. The precedence was not matching https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#operator-precedence.

damieng (approved):

PR #275, opened 2023-01-11
Support `scoped` as identifier name

tamasvajk:
Commit-by-commit review is advised.

This PR is adding support for `scoped` as identifier name. I had to somewhat rework the parameter rule, which previously allowed parameter modifiers without a parameter type. I found it somewhat odd that parameters can be declared without a type, but that's the case in lambdas, such as `System.Func<int, int, object> l = (i, j) => null;`, and also in the rare case of `__arglist`. I don't think parameter modifiers could be used in such cases.

I have no systematic way of finding issues regarding contextual keywords (such as the issue fixed in the [last commit](https://github.com/tree-sitter/tree-sitter-c-sharp/pull/275/commits/21d7dbcc22b7022e6897073b4620254fc97d27bd)), so there might be other cases too, when `scoped` doesn't work as an identifier.

### DIFF
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -135,3 +135,190 @@ file class file { }
     (modifier)
     (identifier)
     (declaration_list)))
+
+================================================================================
+Scoped contextual keyword
+================================================================================
+
+void scoped() { }
+void m(scoped p) { }
+void m(scoped ref int p) { }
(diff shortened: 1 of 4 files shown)


damieng (approved):

tamasvajk:
@damieng I've rebased this PR, and updated the `file-sizes.txt`.

PR #279, opened 2023-01-12
Remove contextual keyword `set`, `get`, `init`, `add`, `remove`

tamasvajk:
`set`, `get`, `init`, `add`, `remove` are not needed in the `_contextual_keywords` list. They can only be used as a property accessor name, which doesn't seem to interfere with identifiers.

I've split the PR into two commits to make displaying the changes in the first commit. (I'm getting "Sorry, this diff is taking too long to generate." when viewing the diff of the entire PR)

### DIFF
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -323,3 +323,93 @@ class scoped { }
   (class_declaration
     (identifier)
     (declaration_list)))
+
+================================================================================
+Set contextual keyword
+================================================================================
+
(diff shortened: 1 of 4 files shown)


damieng (approved):
Just need to rebase and get the file-sizes in but otherwise diff looks good (splitting the parser.c out made it possible to view online, thanks!)

tamasvajk:
I've rebased and squashed. Also added the `file-sizes.txt`.

Issue #163, opened 2021-02-18
Parse error when operating on variable named `var`


Sjord:
```
using System;

public class HelloWorldVar
{
    public static void Main()
    {
        int var = 3;
        Console.WriteLine(var + 1);
    }
}
```

```sh
$ tree-sitter parse Program.cs
...
Program.cs	0 ms	(ERROR [7, 26] - [7, 29])
```

damieng:
Looks like a case where #101 was not comprehensive enough.

initram:
The same issue exists (and is documented in the readme) for async and await. They can not be used as identifiers either.

PR #280, opened 2023-01-12
Add `var` to contextual keywords

tamasvajk:
Fixes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/163.

Adding `var` as a contextual keyword meant that implicit type and name conflicted, and because they could show up in the same position in an AST as a type, dynamic precedence was specified to prefer implicit type. I didn't want to modify the `_name` rule, but only cases when `_name` is used as a type, so I introduced an intermediate wrapper that adds the dynamic precedence.

### DIFF
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -413,3 +413,99 @@ class set { }
   (class_declaration
     (identifier)
     (declaration_list)))
+
+================================================================================
+Var contextual keyword
+================================================================================
+
+void var() { }
+void m(var p) { }
+void m(int var) { }
+void m()
+{
+    var v = null;
+    int var = var;
+    var var = 1;
+
+    var();
+    m(var);
+
+    var x = var + 1;
+}
+
+class var { }
+
+--------------------------------------------------------------------------------
+
+(compilation_unit
+  (global_statement
+    (local_function_statement
+      (predefined_type)
+      (identifier)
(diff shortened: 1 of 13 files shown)


corpus/contextual-keywords.txt
```diff
@@ -135,3 +135,99 @@ file class file { }
...
+            (variable_declarator
+              (identifier)
+              (equals_value_clause
+                (null_literal)))))
+        (local_declaration_statement
+          (variable_declaration
+            (predefined_type)
+            (variable_declarator
+              (identifier)))
+          (ERROR
```
damieng: Looks like we have parsing errors here.

grammar.js
```diff
@@ -78,6 +78,11 @@ module.exports = grammar({
     [$._type, $._pointer_base_type],
     [$._type, $._ref_base_type],
 
+    [$._type_name, $._nullable_base_type],
+    [$._type_name, $._array_base_type],
+    [$._type_name, $._pointer_base_type],
+    [$._type_name, $._ref_base_type],
+
```
damieng, line 85: Can you rebase and run with the line number scripts. I'm getting anxious about all these conflict rules.

tamasvajk: After the rebase, these are not needed. I've pushed a commit with the file sizes, there's basically no change.

damieng (approved):

PR #284, opened 2023-01-12
Add file size checking to prevent large unintended growth

damieng:
This script dumps out the generated file sizes and number of lines into a file we check in so we can at-a-glance see how changes are impacting the size of the generated parser in order to prevent large unintended growth.


damieng:
Hey @tamasvajk do you think this would help with our workflow to keep an eye on generated file sizes?

tamasvajk (commented):
Looks good. 

Some options/questions:

- I was planning on adding a GH action that would check that none of the files increased/decreased by more than a couple of percentages. The check would otherwise fail. It wouldn't be a required workflow, so the failure would only notify us, and not block merging. This way we wouldn't need to commit anything to the repo
- There's nothing currently that would enforce that the file is updated when any file changes in the `src` folder. Should there be an action, which checks if the sizes in the file and on disk actually match?
- Would `wc` report the same values on each machine/platform?

Sjord:
In [tree-sitter-php](https://github.com/tree-sitter/tree-sitter-php/blob/master/.github/pull_request_template.md) they (manually) check the number of conflicts and the parser's state count. Maybe these would also be interesting metrics for parser size.

damieng:
I'm going to add this for now - we could check extra metrics for now. I just want to avoid the major expansion we just had that is now causing LFS warnings and diff timeouts that is making things hard to work with. Compilation is also now very slow.

I think it's worth re-starting that the goal is to have a fast and efficient parser for C# and that may come at the trade off of not supporting every single unusual syntax scenario so we want to be careful about blowing up the parser to support edge cases.

PR #285, opened 2023-01-12
WIP: Restrict the invoked expression in invocation_expression

tamasvajk:


tamasvajk:
This is not the right approach. `var x = a<b>;` is a valid case for a generic name as an expression.

PR #291, opened 2023-01-19
Adds GH tags queries

BekaValentine:
This PR adds new tags queries to `tags.scm` to support code search on Github.com.

### DIFF
--- a/queries/tags.scm
+++ b/queries/tags.scm
@@ -1,46 +1,23 @@
-(class_declaration
- name: (identifier) @name
- ) @definition.class
+(class_declaration name: (identifier) @name) @definition.class
 
-(class_declaration
-   bases: (base_list (_) @name)
- ) @reference.class
+(class_declaration bases: (base_list (_) @name)) @reference.class
 
-(interface_declaration
- name: (identifier) @name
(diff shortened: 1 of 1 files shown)


queries/tags.scm
```diff
@@ -1,46 +1,23 @@
...
+
+(type_constraint type: (identifier) @name) @reference.class
+
+(variable_declaration type: (identifier) @name) @reference.class
+
+(invocation_expression function: (member_access_expression name: (identifier) @name)) @reference.send
+
+(namespace_declaration name: (identifier) @name) @definition.module
+
+(namespace_declaration name: (identifier) @name) @module
```
damieng, line 23: Is this the only new line? Looks same as definition above with `@definition` removed.

What does this do?

BekaValentine: This adds a new tag to namespace declarations, so that now they'll be tagged as both `definition.module` and `module`. We're only adding rather than replacing because we don't want to cause any problems for anyone who might be relying on the `definition.module` tag.

damieng (approved):

damieng:
Thanks for the contribution and explanation.

PR #293, opened 2023-02-10
Fix C# raw_string_literal

gonglinyuan:
Example program:

```csharp
# define DEBUG

using System;



/** <summary>Class constructor</summary>
  * <param>param1</param>
*/
public class HelloWorld {
  /// <summary>
  ///   This is a C sharp docstring.
  /// </summary>
  public static void Main(string[] args) {
    int x, y;
    string z, zz;
    char zzz;
    string tb1, tb2, tb3;
    for (x = 0; x < 10; x++) {
      y = x + 1;
      break;  // 123
      // this is another comment.
    }
    /* This is a block comment,
    that has two lines.
    */
    if (x < 5) {
      y = 2;
    }
#line 100 "special"
    z = "123";
    zz = @"456";
    zzz = 'T';
    tb1 = """text block 1""";
    tb2 = """
    text block 2
    """;
    tb3 = """"
    text block 3
    """";
    Console.WriteLine($"Hello, {z}! Do you know {tb1}?");
    var bytes = "hello"u8;
#if DEBUG
    Console.WriteLine("Debug version");
#endif
  }
}
```

In this program, `tb1`, `tb2` and `tb3` are assigned with three `raw_string_literal`s. However, the parser in tree-sitter considers them as a single `raw_string_literal` of

```
text block 1""";
    tb2 = """
    text block 2
    """;
    tb3 = """"
    text block 3
    "
```

which is incorrect. A change in grammar can fix this issue.

### DIFF
--- a/corpus/literals.txt
+++ b/corpus/literals.txt
@@ -1179,3 +1179,68 @@ var x = $"""The point
                           (identifier)
                           (identifier)))))))
               (interpolated_raw_string_text))))))))
+
+================================================================================
+multiple raw string literals should not be combined
+================================================================================
+
+using System;
+
+public class RawStringLiterals {
+  public static void Main(string[] args) {
+    string tb1, tb2, tb3;
+    tb1 = """text block 1""";
+    tb2 = """
+    text block 2
+    """;
+    tb3 = """"
+    text block 3
+    """";
+  }
+}
+
+--------------------------------------------------------------------------------
+
+(compilation_unit
+  (using_directive
+    (identifier))
+  (class_declaration
+    (modifier)
+    name: (identifier)
+    body: (declaration_list
+      (method_declaration
+        (modifier)
+        (modifier)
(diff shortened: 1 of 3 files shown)


damieng (changes requested):
Hi @gonglinyuan thanks for this, it looks like a good improvement.

Any chance you can add some new snippet tests to one of the corpus/.txt files to test the new scenarios where it failed before - that way we can be sure we don't regress it again in the future.

Let me know if you have questions.

gonglinyuan:
> Hi @gonglinyuan thanks for this, it looks like a good improvement.
> 
> Any chance you can add some new snippet tests to one of the corpus/.txt files to test the new scenarios where it failed before - that way we can be sure we don't regress it again in the future.
> 
> Let me know if you have questions.

I added a test to `literals.txt`

damieng:
Thank-you for fixing this!

PR #294, opened 2023-02-11
Fix "logical and" parsed as "cast + dereference"

gonglinyuan:
The expression `(a) && b` in this statement:

```csharp
bool c = (a) && b;
```

was mistakenly parsed as a cast expression of type `a` applied to `&&b`, which is variable `b` dereferenced via `&` twice. However, it should have been parsed just as a logical_and expression.

This PR changed the grammar to fix this issue.

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -3375,3 +3375,23 @@ var x = nameof(A.B);
                     (identifier)
                     (identifier)))))))))))
 
+================================================================================
+Dereference versus logical and
+================================================================================
+
+bool c = (a) && b;
+
+--------------------------------------------------------------------------------
+
+(compilation_unit
+  (global_statement
+    (local_declaration_statement
+      (variable_declaration
+        type: (predefined_type)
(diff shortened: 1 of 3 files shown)


damieng (approved):

damieng:
Thanks for the fix!

damieng:
I'm going to have to revert this as it's causing the following regression:

```csharp
Sprite mySprite = GetNode<Sprite>("Sprite");
```

Should be an `invocation_expression` on the right-hand side but this change is causing it to be an (incorrect) `binary_expression` now.

gonglinyuan:
> I'm going to have to revert this as it's causing the following regression:
> 
> ```cs
> Sprite mySprite = GetNode<Sprite>("Sprite");
> ```
> 
> Should be an `invocation_expression` on the right-hand side but this change is causing it to be an (incorrect) `binary_expression` now.
> 
> Not sure how we didn't have test coverage for this scenario.

#299 should fix this issue

PR #295, opened 2023-02-14
Extract string literal contents and encoding

hvitved:
The `string_literal_fragment` rule was previously hidden, which means that we cannot access the actual contents of string literals. This PR changes that, and also gives access to the optional UTF8 encoding marker.


hvitved:
@damieng : Thanks for the review. OK to merge?

damieng:
Thanks @hvitved !

PR #296, opened 2023-02-15
Extract character literal contents

hvitved:
Like https://github.com/tree-sitter/tree-sitter-c-sharp/pull/295, but for character literals.


damieng (approved):

damieng:
Thanks!

PR #299, opened 2023-03-03
Fix "logical and" Parsed as "cast + deref" While Preserving Generic Invocations

gonglinyuan:
I am pleased to submit this pull request to resolve the issue mentioned in the comments of #294. While #294 addressed the problem of "logical and" being mistakenly parsed as "cast + dereference", it inadvertently caused invocation expressions with generics to be parsed as binary expressions. So #294 was reverted and the test was removed. A new test about generic invocations was added in the commit fc6be5ae61c8788e0c028e757f132d67bfc538f5

To address this issue, I have added precedence to the generate type argument list `<XXX>` to make it bind more tightly. With this update, the parser now passes both the test added in #294 and the test added in the commit fc6be5ae61c8788e0c028e757f132d67bfc538f5

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -3424,3 +3424,25 @@ MyFunction<A,B>(1);
           (argument_list
             (argument
               (integer_literal)))))))
+
+
+================================================================================
+Dereference versus logical and
+================================================================================
+
+bool c = (a) && b;
+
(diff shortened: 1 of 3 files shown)


damieng (approved):

damieng:
Thanks for tackling this again!

PR #300, opened 2023-04-04
Add some more named fields to the grammar

tamasvajk:
This PR adds some more named fields to the grammar. (I couldn't add `field('attributes', $.attribute_list)` because that significantly increases the parser size.)

Additionally, the PR simplifies the highlighting of types by using the fields named `type` (`(_ type: (identifier) @type)`) instead of defining individually for each construct where types are located (`(identifier) @type` in individual constructs).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -124,12 +124,12 @@ module.exports = grammar({
 
   rules: {
     compilation_unit: $ => seq(
-      repeat($.extern_alias_directive),
-      repeat($.using_directive),
-      repeat($.global_attribute_list),
+      repeat(field('extern_alias', $.extern_alias_directive)),
+      repeat(field('using', $.using_directive)),
+      repeat(field('attributes', $.global_attribute_list)),
       choice(
-        seq(repeat($.global_statement), repeat($._namespace_member_declaration)),
-        $.file_scoped_namespace_declaration
+        seq(repeat(field('statement', $.global_statement)), repeat(field('declaration', $._namespace_member_declaration))),
+        field('declaration', $.file_scoped_namespace_declaration)
       )
     ),
 
@@ -172,20 +172,20 @@ module.exports = grammar({
       $.record_struct_declaration
     ),
 
-    extern_alias_directive: $ => seq('extern', 'alias', $.identifier, ';'),
+    extern_alias_directive: $ => seq('extern', 'alias', field('name', $.identifier), ';'),
 
     using_directive: $ => seq(
       optional('global'),
       'using',
       optional(choice(
         'static',
-        $.name_equals
+        field('alias', $.name_equals)
       )),
-      $._name,
+      field('name', $._name),
       ';'
     ),
 
-    name_equals: $ => prec(1, seq($._identifier_or_global, '=')),
+    name_equals: $ => prec(1, seq(field('name', $._identifier_or_global), '=')),
 
     _name: $ => choice(
       $.alias_qualified_name,
@@ -193,14 +193,17 @@ module.exports = grammar({
       $._simple_name
     ),
 
-    alias_qualified_name: $ => seq($._identifier_or_global, '::', $._simple_name),
+    alias_qualified_name: $ => seq(
+      field('alias', $._identifier_or_global),
+      '::',
+      field('name', $._simple_name)),
 
     _simple_name: $ => choice(
       $.generic_name,
       $._identifier_or_global
     ),
 
-    generic_name: $ => seq($.identifier, $.type_argument_list),
+    generic_name: $ => seq(field('name', $.identifier), field('type_arguments', $.type_argument_list)),
 
     // Intentionally different from Roslyn to avoid non-matching
     // omitted_type_argument in a lot of unnecessary places.
@@ -208,17 +211,20 @@ module.exports = grammar({
       '<',
       choice(
         repeat(','),
-        commaSep1($._type),
+        commaSep1(field('type', $._type)),
       ),
       '>'
     )),
 
-    qualified_name: $ => prec(PREC.DOT, seq($._name, '.', $._simple_name)),
+    qualified_name: $ => prec(PREC.DOT, seq(
+      field('left', $._name),
+      '.',
+      field('right', $._simple_name))),
 
     attribute_list: $ => seq(
       '[',
       optional($.attribute_target_specifier),
-      commaSep1($.attribute),
+      commaSep1(field('attribute', $.attribute)),
       optional(','),
       ']'
     ),
@@ -230,35 +236,35 @@ module.exports = grammar({
 
     attribute: $ => seq(
       field('name', $._type_name),
-      optional($.attribute_argument_list)
+      optional(field('arguments', $.attribute_argument_list))
     ),
 
     attribute_argument_list: $ => seq(
       '(',
-      commaSep($.attribute_argument),
+      commaSep(field('argument', $.attribute_argument)),
       ')'
     ),
 
     attribute_argument: $ => seq(
       optional(choice($.name_equals, $.name_colon)),
-      $._expression
+      field('expression', $._expression)
     ),
 
     global_attribute_list: $ => seq(
       '[',
       choice('assembly', 'module'),
       ':',
-      commaSep($.attribute),
+      commaSep(field('attribute', $.attribute)),
       ']'
     ),
 
-    name_colon: $ => seq($._identifier_or_global, ':'),
+    name_colon: $ => seq(field('name', $._identifier_or_global), ':'),
 
     event_field_declaration: $ => prec.dynamic(1, seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       'event',
-      $.variable_declaration,
+      field('declaration', $.variable_declaration),
       ';'
     )),
 
@@ -289,48 +295,48 @@ module.exports = grammar({
 
     variable_declaration: $ => seq(
       field('type', $._type),
-      commaSep1($.variable_declarator)
+      commaSep1(field('variable_declarator', $.variable_declarator))
     ),
 
     variable_declarator: $ => seq(
-      choice($.identifier, $.tuple_pattern),
-      optional($.bracketed_argument_list),
-      optional($.equals_value_clause)
+      choice(field('name', $.identifier), $.tuple_pattern),
+      optional(field('arguments', $.bracketed_argument_list)),
+      field('initializer', optional($.equals_value_clause))
     ),
 
     bracketed_argument_list: $ => seq(
       '[',
-      commaSep1($.argument),
+      commaSep1(field('argument', $.argument)),
       ']'
     ),
 
     tuple_pattern: $ => seq(
       '(',
-      commaSep1(choice($.identifier, $.discard, $.tuple_pattern)),
+      commaSep1(choice(field('name', $.identifier), $.discard, $.tuple_pattern)),
       ')'
     ),
 
     argument: $ => prec(1, seq(
       optional($.name_colon),
       optional(choice('ref', 'out', 'in')),
-      choice(
+      field('expression', choice(
         $._expression,
         $.declaration_expression
-      )
+      ))
     )),
 
-    equals_value_clause: $ => seq('=', $._expression),
+    equals_value_clause: $ => seq('=', field('value', $._expression)),
 
     field_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
-      $.variable_declaration,
+      repeat(field('modifier', $.modifier)),
+      field('declaration', $.variable_declaration),
       ';'
     ),
 
     constructor_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       field('name', $.identifier),
       field('parameters', $.parameter_list),
       optional($.constructor_initializer),
@@ -349,10 +355,10 @@ module.exports = grammar({
 
     implicit_parameter: $ => field('name', $.identifier),
 
-    _formal_parameter_list: $ => commaSep1(choice(
+    _formal_parameter_list: $ => commaSep1(field('parameter', choice(
       $.parameter,
       $._parameter_array
-    )),
+    ))),
 
     _parameter_type_with_modifiers: $ => seq(
       alias(optional('this'), $.parameter_modifier),
@@ -365,7 +371,7 @@ module.exports = grammar({
       repeat($.attribute_list),
       optional($._parameter_type_with_modifiers),
       field('name', $.identifier),
-      optional($.equals_value_clause)
+      optional(field('default', $.equals_value_clause))
     ),
 
     parameter_modifier: $ => choice('ref', 'out', 'this', 'in', 'scoped'),
@@ -373,25 +379,25 @@ module.exports = grammar({
     _parameter_array: $ => seq(
       repeat($.attribute_list),
       'params',
-      choice($.array_type, $.nullable_type),
+      field('type', choice($.array_type, $.nullable_type)),
       field('name', $.identifier),
     ),
 
     constructor_initializer: $ => seq(
       ':',
       choice('base', 'this'),
-      $.argument_list
+      field('arguments', $.argument_list)
     ),
 
-    argument_list: $ => seq('(', commaSep($.argument), ')'),
+    argument_list: $ => seq('(', commaSep(field('argument', $.argument)), ')'),
 
-    block: $ => seq('{', repeat($._statement), '}'),
+    block: $ => seq('{', repeat(field('statement', $._statement)), '}'),
 
-    arrow_expression_clause: $ => seq('=>', $._expression),
+    arrow_expression_clause: $ => seq('=>', field('expression', $._expression)),
 
     conversion_operator_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       choice(
         'implicit',
         'explicit'
@@ -421,7 +427,7 @@ module.exports = grammar({
 
     method_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       field('type', $._type),
       optional($.explicit_interface_specifier),
       field('name', $.identifier),
@@ -433,7 +439,7 @@ module.exports = grammar({
 
     explicit_interface_specifier: $ => prec(PREC.DOT, seq($._name, '.')),
 
-    type_parameter_list: $ => seq('<', commaSep1($.type_parameter), '>'),
+    type_parameter_list: $ => seq('<', commaSep1(field('type_parameter', $.type_parameter)), '>'),
 
     type_parameter: $ => seq(
       repeat($.attribute_list),
@@ -463,7 +469,7 @@ module.exports = grammar({
 
     operator_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       field('type', $._type),
       optional($.explicit_interface_specifier),
       'operator',
@@ -492,7 +498,7 @@ module.exports = grammar({
 
     event_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       'event',
       field('type', $._type),
       optional($.explicit_interface_specifier),
@@ -505,20 +511,20 @@ module.exports = grammar({
 
     accessor_list: $ => seq(
       '{',
-      repeat($.accessor_declaration),
+      repeat(field('accessor', $.accessor_declaration)),
       '}'
     ),
 
     accessor_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       choice('get', 'set', 'add', 'remove', 'init', $.identifier),
       $._function_body
     ),
 
     indexer_declaration: $ => seq(
       repeat($.attribute_list),
-      repeat($.modifier),
+      repeat(field('modifier', $.modifier)),
       field('type', $._type),
       optional($.explicit_interface_specifier),
       'this',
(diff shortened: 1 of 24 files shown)


script/file_sizes.txt
```diff
@@ -1,5 +1,5 @@
-src/grammar.json    	0.2MB	     10984
-src/node-types.json 	0.1MB	      7677
-src/parser.c        	49.4MB	   1553316
+src/grammar.json    	0.3MB	     11864
+src/node-types.json 	0.1MB	      8495
+src/parser.c        	58.6MB	   1848207
 src/scanner.c       	0.0MB	        29
-total               	49.8MB	   1572006
+total               	59.0MB	   1868595
```
tamasvajk: I find it rather odd that adding field names has such a big impact on the parser size. @damieng do you have some insight why this is the case?

hvitved: @maxbrunsfeld may be able to answer this as well.

tamasvajk:
Doing the opposite and removing all `field()` calls from the grammar has significant impact on the parser size: 
```diff
--- a/script/file_sizes.txt
+++ b/script/file_sizes.txt
@@ -1,5 +1,5 @@
-src/grammar.json       0.2MB        10954
-src/node-types.json    0.1MB         7580
-src/parser.c           47.9MB     1505318
+src/grammar.json       0.2MB        10098
+src/node-types.json    0.1MB         6307
+src/parser.c           38.8MB     1221658
 src/scanner.c          0.0MB           37
-total                  48.3MB     1523889
+total                  39.1MB     1238100
```

damieng:
Doing that though will break a lot of the usefulness of the project and break GitHub semantic use no?

tamasvajk:
> Doing that though will break a lot of the usefulness of the project and break GitHub semantic use no?

Yes, sure. I'm not saying we should do this. Only highlighting that field names contribute significantly to the parser size.

damieng:
Yeah it's weird, I didn't imagine they would add much at all to the size.

tamasvajk:
> Yeah it's weird, I didn't imagine they would add much at all to the size.

It looks like not all field names cause issues: https://github.com/tree-sitter/tree-sitter-c-sharp/pull/314.

grammar.js
```diff
@@ -725,9 +731,9 @@ module.exports = grammar({
 
     // grammar.txt marks this non-optional and includes omitted_array_size_expression in
     // expression but we can't match empty rules.
-    array_rank_specifier: $ => seq('[', commaSep(optional($._expression)), ']'),
+    array_rank_specifier: $ => seq('[', commaSep(optional(field('expression', $._expression))), ']'),
 
-    nullable_type: $ => seq($._nullable_base_type, '?'),
+    nullable_type: $ => seq(field('element_type', $._nullable_base_type), '?'),
```
hvitved, line 736: `underlying_type`?

hvitved (commented):
LGTM, a few comments.

grammar.js
```diff
@@ -709,7 +715,7 @@ module.exports = grammar({
       'var'),
 
     array_type: $ => seq(
-      field('type', $._array_base_type),
+      field('element_type', $._array_base_type),
```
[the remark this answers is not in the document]
tamasvajk: I don't know either if breaking changes are okay. At the same time, `element_type` matches `ElementType`, which is used by Roslyn.

grammar.js
tamasvajk: Roslyn uses `ElementType` in `int?`.

grammar.js
```diff
@@ -736,7 +742,7 @@ module.exports = grammar({
       $.tuple_type
     ),
 
-    pointer_type: $ => seq($._pointer_base_type, '*'),
+    pointer_type: $ => seq(field('element_type', $._pointer_base_type), '*'),
```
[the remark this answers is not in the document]
tamasvajk: Same comment as above, Roslyn uses `ElementType` in `int*`.

grammar.js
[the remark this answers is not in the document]
damieng: Will GitHub Semantic know that element_type is a type?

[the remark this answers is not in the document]
hendrikvanantwerpen: In general, breaking changes should be possible if there's a good case. In this case, it doesn't seem to improve anything for downstream processing, so my preference would be to stick with what we have. These kind of changes do break all queries for this node written against this grammar, after all.

grammar.js
```diff
@@ -725,9 +731,9 @@ module.exports = grammar({
 
     // grammar.txt marks this non-optional and includes omitted_array_size_expression in
     // expression but we can't match empty rules.
-    array_rank_specifier: $ => seq('[', commaSep(optional($._expression)), ']'),
+    array_rank_specifier: $ => seq('[', commaSep(optional(field('expression', $._expression))), ']'),
```
hendrikvanantwerpen, line 734: As above, can we drop the element field?

grammar.js
hendrikvanantwerpen: My preference would be to stick with `type`, in line with other places in the grammar.

grammar.js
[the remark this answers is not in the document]
hendrikvanantwerpen: Idem, I'd stick with `type`.

grammar.js
```diff
@@ -834,7 +840,7 @@ module.exports = grammar({
 
     tuple_type: $ => seq(
       '(',
-      commaSep2($.tuple_element),
+      commaSep2(field('element', $.tuple_element)),
```
hendrikvanantwerpen, line 843: As above, can we drop the field since this is a simple list?

grammar.js
```diff
@@ -876,18 +882,18 @@ module.exports = grammar({
...
 
-    do_statement: $ => seq('do', $._statement, 'while', '(', $._expression, ')', ';'),
+    do_statement: $ => seq('do', field('statement', $._statement), 'while', '(', field('condition', $._expression), ')', ';'),
 
     empty_statement: $ => ';',
 
     expression_statement: $ => seq($._expression_statement_expression, ';'),
 
-    fixed_statement: $ => seq('fixed', '(', $.variable_declaration, ')', $._statement),
+    fixed_statement: $ => seq('fixed', '(', field('declaration', $.variable_declaration), ')', $._statement),
```
hendrikvanantwerpen, line 891: Also introduce a field `statement`?

grammar.js
```diff
@@ -876,18 +882,18 @@ module.exports = grammar({
...
 
     for_statement: $ => seq(
       'for',
       '(',
-      field('initializer', optional(choice($.variable_declaration, commaSep1($._expression)))),
+      field('initializer', optional(choice(field('declaration', $.variable_declaration), commaSep1($._expression)))),
```
hendrikvanantwerpen, line 896: I am not sure what the effect of nesting is here. Perhaps something like this is less ambiguous:

```suggestion
      optional(choice(field('declaration', $.variable_declaration), commaSep1(field('initializer', $._expression))))),
```

grammar.js
```diff
@@ -958,9 +964,9 @@ module.exports = grammar({
       $._function_body
     ),
 
-    lock_statement: $ => seq('lock', '(', $._expression, ')', $._statement),
+    lock_statement: $ => seq('lock', '(', field('expression', $._expression), ')', $._statement),
```
hendrikvanantwerpen, line 967: As above, `value`? Perhaps also name the statement.

grammar.js
```diff
@@ -958,9 +964,9 @@ module.exports = grammar({
...
 
-    return_statement: $ => seq('return', optional($._expression), ';'),
+    return_statement: $ => seq('return', optional(field('expression', $._expression)), ';'),
```
hendrikvanantwerpen, line 969: As before, `value`?

grammar.js
```diff
@@ -1025,10 +1031,10 @@ module.exports = grammar({
...
 
     relational_pattern: $ => prec.left(choice(
-      seq('<', $._expression),
-      seq('<=', $._expression),
-      seq('>', $._expression),
-      seq('>=', $._expression)
+      seq('<', field('expression', $._expression)),
+      seq('<=', field('expression', $._expression)),
+      seq('>', field('expression', $._expression)),
+      seq('>=', field('expression', $._expression))
```
hendrikvanantwerpen, line 1037: As above, `value`?

grammar.js
```diff
@@ -1101,7 +1107,7 @@ module.exports = grammar({
       ')',
     )),
 
-    expression_colon: $ => seq($._expression, ':'),
+    expression_colon: $ => seq(field('expression', $._expression), ':'),
```
hendrikvanantwerpen, line 1110: Idem

grammar.js
```diff
@@ -1117,13 +1123,13 @@ module.exports = grammar({
 
     var_pattern: $ => prec(1, seq('var', $._variable_designation)),
 
-    when_clause: $ => seq('when', $._expression),
+    when_clause: $ => seq('when', field('expression', $._expression)),
```
hendrikvanantwerpen, line 1126: Idem

grammar.js
```diff
@@ -1117,13 +1123,13 @@ module.exports = grammar({
...
 
-    case_switch_label: $ => prec.left(1, seq('case', $._expression, ':')),
+    case_switch_label: $ => prec.left(1, seq('case', field('expression', $._expression), ':')),
```
hendrikvanantwerpen, line 1128: Idem

grammar.js
```diff
@@ -1117,13 +1123,13 @@ module.exports = grammar({
...
 
     default_switch_label: $ => prec.left(1, seq('default', ':')),
 
-    throw_statement: $ => seq('throw', optional($._expression), ';'),
+    throw_statement: $ => seq('throw', optional(field('expression', $._expression)), ';'),
```
hendrikvanantwerpen, line 1132: Idem

grammar.js
```diff
@@ -1156,17 +1162,17 @@ module.exports = grammar({
       optional('await'),
       'using',
       '(',
-      choice($.variable_declaration, $._expression),
+      choice(field('declaration', $.variable_declaration), field('expression', $._expression)),
```
hendrikvanantwerpen, line 1165: Idem

grammar.js
```diff
@@ -1156,17 +1162,17 @@ module.exports = grammar({
...
     ),
 
-    while_statement: $ => seq('while', '(', $._expression, ')', $._statement),
+    while_statement: $ => seq('while', '(', field('condition', $._expression), ')', field('statement', $._statement)),
 
     yield_statement: $ => seq(
       'yield',
       choice( // grammar.txt incorrectly allows "break expression", we do not.
-        seq('return', $._expression),
+        seq('return', field('expression', $._expression)),
```
hendrikvanantwerpen, line 1175: Idem

grammar.js
```diff
@@ -1209,26 +1215,26 @@ module.exports = grammar({
...
-      $.array_type,
-      optional($.initializer_expression)
+      field('type', $.array_type),
+      field('initializer', optional($.initializer_expression))
     )),
 
     initializer_expression: $ => seq(
       '{',
-      commaSep($._expression),
+      commaSep(field('expression', $._expression)),
```
hendrikvanantwerpen, line 1224: Again a simple list, so perhaps drop the field? Otherwise `value`.

grammar.js
```diff
@@ -1209,26 +1215,26 @@ module.exports = grammar({
...
       field('left', $._lvalue_expression),
-      $.assignment_operator,
+      field('operator', $.assignment_operator),
       field('right', $._expression)
     )),
 
     assignment_operator: $ => choice('=', '+=', '-=', '*=', '/=', '%=', '&=', '^=', '|=', '<<=', '>>=', '>>>=', '??='),
 
-    await_expression: $ => prec.right(PREC.UNARY, seq('await', $._expression)),
+    await_expression: $ => prec.right(PREC.UNARY, seq('await', field('expression', $._expression))),
```
hendrikvanantwerpen, line 1237: Also `value`

grammar.js
```diff
@@ -1238,8 +1244,8 @@ module.exports = grammar({
     ))),
 
     checked_expression: $ => choice(
-      seq('checked', '(', $._expression, ')'),
-      seq('unchecked', '(', $._expression, ')')
+      seq('checked', '(', field('expression', $._expression), ')'),
+      seq('unchecked', '(', field('expression', $._expression), ')')
```
hendrikvanantwerpen, line 1248: Idem

grammar.js
```diff
@@ -1275,21 +1281,21 @@ module.exports = grammar({
       field('subscript', $.bracketed_argument_list)
     )),
 
-    element_binding_expression: $ => $.bracketed_argument_list,
+    element_binding_expression: $ => field('subscript', $.bracketed_argument_list),
```
hendrikvanantwerpen, line 1284: Perhaps `arguments` is more in line with existing field names?

grammar.js
```diff
@@ -1275,21 +1281,21 @@ module.exports = grammar({
...
 
     implicit_array_creation_expression: $ => seq(
       'new',
       '[',
       repeat(','),
       ']',
-      $.initializer_expression
+      field('initializer', $.initializer_expression)
```
hendrikvanantwerpen, line 1291: This is a maybe, but it could also just be `value`

grammar.js
```diff
@@ -1275,21 +1281,21 @@ module.exports = grammar({
...
     ),
 
     implicit_stack_alloc_array_creation_expression: $ => seq(
       'stackalloc',
       '[',
       ']',
-      $.initializer_expression
+      field('initializer', $.initializer_expression)
```
hendrikvanantwerpen, line 1298: Idem

grammar.js
```diff
@@ -1399,9 +1405,9 @@ module.exports = grammar({
     _parenthesized_lvalue_expression: $ => seq('(', $._lvalue_expression, ')'),
 
     postfix_unary_expression: $ => prec.left(PREC.POSTFIX, choice(
-      seq($._expression, '++'),
-      seq($._expression, '--'),
-      seq($._expression, '!')
+      seq(field('expression', $._expression), '++'),
+      seq(field('expression', $._expression), '--'),
+      seq(field('expression', $._expression), '!')
```
hendrikvanantwerpen, line 1410: Idem

grammar.js
```diff
@@ -1415,18 +1421,18 @@ module.exports = grammar({
         '--',
         '^',
         '~'
-      ].map(operator => seq(operator, $._expression)))),
+      ].map(operator => seq(operator, field('expression', $._expression))))),
```
hendrikvanantwerpen, line 1424: Idem

grammar.js
```diff
@@ -1415,18 +1421,18 @@ module.exports = grammar({
...
 
-    _pointer_indirection_expression: $ => prec.right(PREC.UNARY, seq('*', $._expression)),
+    _pointer_indirection_expression: $ => prec.right(PREC.UNARY, seq('*', field('expression', $._expression))),
```
hendrikvanantwerpen, line 1426: This production (because it starts with `_`) doesn't introduce an intermediate node, so I don't think it should introduce a field at all. That is, assuming a field is already created at the use site...

grammar.js
```diff
@@ -1415,18 +1421,18 @@ module.exports = grammar({
...
     query_expression: $ => seq($.from_clause, $._query_body),
 
     from_clause: $ => seq(
       'from',
-      optional($._type),
+      optional(field('type', $._type)),
       $.identifier,
       'in',
-      $._expression
+      field('in_expression', $._expression)
```
hendrikvanantwerpen, line 1435: As above, `value`

grammar.js
```diff
@@ -1415,18 +1421,18 @@ module.exports = grammar({
...
-    _pointer_indirection_expression: $ => prec.right(PREC.UNARY, seq('*', $._expression)),
+    _pointer_indirection_expression: $ => prec.right(PREC.UNARY, seq('*', field('expression', $._expression))),
 
     query_expression: $ => seq($.from_clause, $._query_body),
 
     from_clause: $ => seq(
       'from',
-      optional($._type),
+      optional(field('type', $._type)),
       $.identifier,
```
hendrikvanantwerpen, line 1433: Should we also name this?

grammar.js
```diff
@@ -1520,18 +1526,18 @@ module.exports = grammar({
...
       'stackalloc',
-      $.array_type,
-      optional($.initializer_expression)
+      field('type', $.array_type),
+      field('initializer', optional($.initializer_expression))
     ),
 
     switch_expression: $ => prec(PREC.SWITCH, seq(
-      $._expression,
+      field('expression', $._expression),
```
hendrikvanantwerpen, line 1540: `value`?

grammar.js
```diff
@@ -1484,27 +1490,27 @@ module.exports = grammar({
...
     )),
 
-    ref_expression: $ => prec.right(seq('ref', $._expression)),
+    ref_expression: $ => prec.right(seq('ref', field('expression', $._expression))),
 
     ref_type_expression: $ => seq(
       '__reftype',
       '(',
-      $._expression,
+      field('expression', $._expression),
```
hendrikvanantwerpen, line 1513: `value`?

grammar.js
```diff
@@ -1484,27 +1490,27 @@ module.exports = grammar({
...
       'group',
-      $._expression,
+      field('group_expression', $._expression),
       'by',
-      $._expression
+      field('by_expression', $._expression)
     )),
 
-    select_clause: $ => prec.right(PREC.SELECT, seq('select', $._expression)),
+    select_clause: $ => prec.right(PREC.SELECT, seq('select', field('expression', $._expression))),
```
hendrikvanantwerpen, line 1498: `value`?

grammar.js
```diff
@@ -1484,27 +1490,27 @@ module.exports = grammar({
 
     group_clause: $ => prec.right(PREC.SELECT, seq(
       'group',
-      $._expression,
+      field('group_expression', $._expression),
       'by',
-      $._expression
+      field('by_expression', $._expression)
```
hendrikvanantwerpen, line 1495: As above, `*_value`

grammar.js
```diff
@@ -1471,11 +1477,11 @@ module.exports = grammar({
     ),
 
     _ordering: $ => seq(
-      $._expression,
+      field('expression', $._expression),
       optional(choice('ascending', 'descending'))
     ),
 
-    where_clause: $ => seq('where', $._expression),
+    where_clause: $ => seq('where', field('expression', $._expression)),
```
hendrikvanantwerpen, line 1484: Idem

grammar.js
```diff
@@ -1462,7 +1468,7 @@ module.exports = grammar({
       'let',
       $.identifier,
       '=',
-      $._expression
+      field('expression', $._expression)
```
hendrikvanantwerpen, line 1471: Idem

hendrikvanantwerpen (changes requested):
Thanks for taking the time to introduce these fields, that is really useful for downstream processing of the parse trees!

I have some remarks:

- Existing fields for expressions use `value` for the field name. I would suggest sticking with that convention instead of using `expression`. I feel a bit bad, because when marking them I realized there are many places that needs changing, but I think being consistent is important for long-term maintainability.

- For simple lists (where all the elements are form the same production), I suggest to drop the field name. It doesn't add much, because there are not different fields to distinguish, and it makes the trees very verbose to have them.

tamasvajk:
@hendrikvanantwerpen Thanks for the review.

> Existing fields for expressions use value for the field name. I would suggest sticking with that convention instead of using expression. I feel a bit bad, because when marking them I realized there are many places that needs changing, but I think being consistent is important for long-term maintainability.

No worries that there are many places that need changing, as you said it's important for maintainability. Before I change them, I'll raise one additional concern: I think we're being consistent with the current `expression` vs `value` distinctions; we're following what Roslyn is doing. (I would need to check that I fully matched Roslyn everywhere) For example Roslyn uses `Value` for `6` and `Expression` for 5 in the below example. See in the Syntax Tree view [here](https://sharplab.io/#v2:C4LghgzsA0AmIGoA+BtAggCgJYAIC8OArAJQC6AsAFBgBGUATmAMbA4ACALDgLLYB2rAFb4cANmIBuIA).
```C#
[A(i = 5)]
abstract void M(int j = 6);
```

grammar.js
```diff
@@ -289,48 +295,48 @@ module.exports = grammar({
 
     variable_declaration: $ => seq(
       field('type', $._type),
-      commaSep1($.variable_declarator)
+      commaSep1(field('variable_declarator', $.variable_declarator))
```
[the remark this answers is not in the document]
tamasvajk: This is in a `commaSep1`, shouldn't I instead remove the field name altogether based on your second suggestion?

damieng:
The only reason I followed the Roslyn grammar was to make changes in future C# easier as we can follow what rules change.

If however that's getting in the way of size/usability we should deviate.

hendrikvanantwerpen:
I agree that it is valuable to follow the Roslyn grammar of feasible. Perhaps it makes sense to call this out in the readme, and make it a bit more official? Wording along @damieng's comment above should be fine, I think

grammar.js
tamasvajk: This matches the above `field('subscript', $.bracketed_argument_list)`. For the time being, I'm removing this field name, we can revisit it later...

grammar.js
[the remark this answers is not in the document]
hendrikvanantwerpen: I think this is a slightly different case because there are also _other_ fields here. Since there's also the `type` field, having the field names helps grabbing the right subterms when querying.

On the other hand, for a production similar to

    $.foo = seq('(', commaSep1(field('bar', $.bar)), ')')

all the subterms will have field `bar`. Matching `(foo bar:(_)@bar)` and `(foo (_)@bar)` will have the same result. Adding the field name does not help in distinguishing different subterms of `foo`. In those cases, I suggest to omit it.

Hope that clarifies what I mean!

tamasvajk:
Thanks for all the feedback on this PR. I think it might be easier if we do this step-by-step. In the first step I'm introducing only the fields that have visible impact on some use-cases, such as highlighting. So I'm leaving the fields with `field('type'`. I also renamed `field('element_type'` to `field('type'`, which actually results in some highlighting improvement: https://github.com/tree-sitter/tree-sitter-c-sharp/pull/316

tamasvajk:
Here's another followup PR to add fields to name-like constructs: https://github.com/tree-sitter/tree-sitter-c-sharp/pull/318

PR #306, opened 2023-05-10
Fix unhandled parameter modifier combination

tamasvajk:
Parameter modifiers (`ref`, `in`) can also be applied to extension method parameters (`this`).


damieng (approved):
Looks good!

PR #314, opened 2023-06-15
Change `parameters` field name in lambda expression

tamasvajk:
Changing the below:
```diff
--- a/grammar.js
+++ b/grammar.js
@@ -1178,7 +1178,7 @@ module.exports = grammar({
       repeat($.attribute_list),
       optional(alias(choice('async', 'static', seq('async', 'static'), seq('static', 'async')), $.modifier)),
       optional($._type),
-      choice(field('parameters', $.parameter_list), $.identifier),
+      field('parameters', choice($.parameter_list, $.identifier)),
       '=>',
       field('body', choice($.block, $._expression))
     )),
```
results in a significant parser size reduction:
```diff
--- a/script/file_sizes.txt
+++ b/script/file_sizes.txt
@@ -1,5 +1,5 @@
 src/grammar.json       0.2MB        10975
-src/node-types.json    0.1MB         7685
-src/parser.c           48.9MB     1534188
+src/node-types.json    0.1MB         7689
+src/parser.c           41.2MB     1300296
 src/scanner.c          0.0MB           37
-total                  49.3MB     1552885
+total                  41.5MB     1318997
```

### DIFF
--- a/corpus/attributes.txt
+++ b/corpus/attributes.txt
@@ -10,12 +10,12 @@ Global attributes
 (compilation_unit
   (global_attribute_list
     (attribute
-      (identifier)))
+      name: (identifier)))
   (global_attribute_list
     (attribute
-      (identifier))
+      name: (identifier))
     (attribute
-      (identifier)
+      name: (identifier)
       (attribute_argument_list))))
 
 ================================================================================
@@ -31,14 +31,14 @@ class D {}
   (class_declaration
     (attribute_list
       (attribute
-        (identifier)
+        name: (identifier)
         (attribute_argument_list
           (attribute_argument
             (member_access_expression
-              (identifier)
-              (identifier))))))
-    (identifier)
-    (declaration_list)))
+              expression: (identifier)
+              name: (identifier))))))
+    name: (identifier)
+    body: (declaration_list)))
 
 ================================================================================
 Attributes with qualified name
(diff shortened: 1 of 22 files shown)


tamasvajk:
The output is slightly changed by this modification, which might or might not be acceptable. @damieng what do you think?
```diff
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -277,7 +277,7 @@ class scoped { }
               (identifier)
               (equals_value_clause
                 (lambda_expression
-                  (identifier)
+                  parameters: (identifier)
                   body: (null_literal))))))
         (local_declaration_statement
           (variable_declaration
```

I think it could be improved slightly by making it more similar to:
```
                (lambda_expression
                  parameters: (parameter_list
                    (parameter
                      type: (identifier)
                      name: (identifier)))
```
So something along these lines:
```
                (lambda_expression
                  parameters: (parameter_list
                    (implicitly_typed_parameter
                      name: (identifier)))
```

tamasvajk:
Actually the declaration of `parameter` is:
```
    parameter: $ => seq(
      repeat($.attribute_list),
      optional($._parameter_type_with_modifiers),
      field('name', $.identifier),
      optional($.equals_value_clause)
    ),
```
So everything is optional except the name, so we wouldn't need `implicitly_typed_parameter`.

So in the end, I'd go with something like this:
```
                (lambda_expression
                  parameters: (implicit_parameter_list
                    (parameter
                      name: (identifier)))
```
What do you think?

damieng:
This looks like an improvement all round - wonder why that one has such an impact on the size.

tamasvajk:
> This looks like an improvement all round - wonder why that one has such an impact on the size.

I have no clue yet. I'll have a look at other `field` calls inside a `choice`. 
I pushed some extra commits to adjust the trees:
```diff
--- a/corpus/contextual-keywords.txt
+++ b/corpus/contextual-keywords.txt
@@ -277,7 +277,9 @@ class scoped { }
               (identifier)
               (equals_value_clause
                 (lambda_expression
-                  parameters: (identifier)
+                  parameters: (implicit_parameter_list
+                    (parameter
+                      name: (identifier)))
                   body: (null_literal))))))
         (local_declaration_statement
           (variable_declaration
```

damieng (approved):

PR #324, opened 2023-11-29
Support for do and while loops named fields

GeekMasher:
This pull request added support for the `body` and `condition` named fields for both `do` and `while` loops. This will make it easier to identify which field is which.

### DIFF
--- a/corpus/statements.txt
+++ b/corpus/statements.txt
@@ -67,8 +67,8 @@ class A {
         parameters: (parameter_list)
         body: (block
           (while_statement
-            (boolean_literal)
-            (break_statement)))))))
+            condition: (boolean_literal)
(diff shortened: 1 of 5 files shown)


damieng:
Hi there, thanks for the PR.

It looks like the tests weren't updated to reflect the changes?

`npm run test` will run those tests and show you how the output has changed. If it looks right then you can update the text file to match.

(You can do `npm run test-windows` if you just want to run the corpus tests without the full CI parsing of third-party projects test suite)

GeekMasher:
Thanks @damieng, let me do this now

GeekMasher:
Should be fixed now @damieng, let me know if you need me to update anything else

damieng:
Thanks for your contribution!

Issue #332, opened 2024-04-18
Support primary constructor


juli1:
The current grammar does not support [primary constructors](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors). We should support primary constructors.

neubaner:
I have been working on primary constructors support in my [fork](https://github.com/neubaner/tree-sitter-c-sharp/tree/primary-constructors). Still WIP and there's a lot of tests missing, but hopefully I can submit a PR soon!

Issue #326, opened 2024-03-12
Using statement with a function call with parameters is not parsed as invocation, instead a variable declaration


Sveder:
I'm testing some parsing with big projects and in two instances I think I found the same parsing error. It also reproduces in the TreeSitter playground for csharp. 

Simplified issue:
```
using (CreateAndOpenSelfHost(null, configuration))
{
.....
}
```
parses into a using_statement->variable_declaration
```
              [using_statement](https://tree-sitter.github.io/tree-sitter/playground#) [139, 12] - [151, 13]
                [variable_declaration](https://tree-sitter.github.io/tree-sitter/playground#) [139, 19] - [139, 61]
                  type: [identifier](https://tree-sitter.github.io/tree-sitter/playground#) [139, 19] - [139, 40]
                  [variable_declarator](https://tree-sitter.github.io/tree-sitter/playground#) [139, 40] - [139, 61]
                    [tuple_pattern](https://tree-sitter.github.io/tree-sitter/playground#) [139, 40] - [139, 61]
                      name: [identifier](https://tree-sitter.github.io/tree-sitter/playground#) [139, 41] - [139, 45]
                      name: [identifier](https://tree-sitter.github.io/tree-sitter/playground#) [139, 47] - [139, 60]
```

removing the parameters to the function results in the correct using_statement -> invocation_expression. 
Example code:
```
using (CreateAndOpenSelfHost())
{
.....
}
```
parses into:
```
             [using_statement](https://tree-sitter.github.io/tree-sitter/playground#) [139, 12] - [151, 13]
                [invocation_expression](https://tree-sitter.github.io/tree-sitter/playground#) [139, 19] - [139, 42]
                  function: [identifier](https://tree-sitter.github.io/tree-sitter/playground#) [139, 19] - [139, 40]
                  arguments: [argument_list](https://tree-sitter.github.io/tree-sitter/playground#) [139, 40] - [139, 42]
```

Actual file to parse/copy paste into playground:
https://github.com/NancyFx/Nancy/blob/master/test/Nancy.Hosting.Self.Tests/NancySelfHostFixture.cs#L80
This file has a few using statements - some without arguments that parse to an invocation (as they should) and some, like I linked, with arguments that is parsed to a variable_declarator.

I've started looking at the grammar.js and C# spec to find the issue, it might just be a precedence issues, but wanted to open this issue to ask:
1. Am I right that this is an error?
2. Any advice on how to fix?

Thanks :)

[...]

Issue #319, opened 2023-07-04
Parser errors on malformed C# code different in playground


nibab:
While the following malformed code
`using System; class Program { static void Main() { Console.WriteLine(\"Hello, World); } }`
works fine in the playground (ie the error occurs much lower in the tree) https://tree-sitter.github.io/tree-sitter/playground. see:

```
compilation_unit [0, 0] - [1, 0]
  using_directive [0, 0] - [0, 13]
    identifier [0, 6] - [0, 12]
  class_declaration [0, 14] - [0, 89]
    name: identifier [0, 20] - [0, 27]
    body: declaration_list [0, 28] - [0, 89]
      method_declaration [0, 30] - [0, 87]
        modifier [0, 30] - [0, 36]
        type: void_keyword [0, 37] - [0, 41]
        name: identifier [0, 42] - [0, 46]
        parameters: parameter_list [0, 46] - [0, 48]
        body: block [0, 49] - [0, 87]
          expression_statement [0, 51] - [0, 85]
            invocation_expression [0, 51] - [0, 84]
              function: member_access_expression [0, 51] - [0, 68]
                expression: identifier [0, 51] - [0, 58]
                name: identifier [0, 59] - [0, 68]
              arguments: argument_list [0, 68] - [0, 84]
                ERROR [0, 69] - [0, 71]
                  escape_sequence [0, 69] - [0, 71]
                argument [0, 71] - [0, 76]
                  identifier [0, 71] - [0, 76]
                argument [0, 78] - [0, 83]
                  identifier [0, 78] - [0, 83]

```
this is what the grammar file from tree-sitter-c-sharp-0.20.0 outputs (ie error parsing the method definition):

```
compilation_unit: [0] - [88]
  using_directive: [0] - [13]
    using: [0] - [5]
    identifier: [6] - [12]
    ;: [12] - [13]
  ERROR: [14] - [88]
    class: [14] - [19]
    identifier: [20] - [27]
    {: [28] - [29]
    modifier: [30] - [36]
      static: [30] - [36]
    predefined_type: [37] - [41]
    identifier: [42] - [46]
    parameter_list: [46] - [48]
      (: [46] - [47]
      ): [47] - [48]
    {: [49] - [50]
    member_access_expression: [51] - [68]
      identifier: [51] - [58]
      .: [58] - [59]
      identifier: [59] - [68]
    (: [68] - [69]
    ": [69] - [70]
    string_literal_fragment: [70] - [88]
[...]

PR #333, opened 2024-04-30
Rewrite the grammar

amaanq:
- Reduces state count (tbh, I had it greatly reduced to around ~7-8k, but a few grievances grew it back to nearly what it was before, ~12k, further optimizations can come once I dig more later)
- Fully implements C# 12 and some of 13, including primary constructors, ref readonly parameters, and aliasing *any* type.
- Fixes parsing bugs related to interpolated strings when more than one `$` is used by delegating this to the external scanner for statefulness by leveraging a stack of the current interpolation's info (yes, nested interpolations are handled nicely, no, it was not fun at times to implement)
- Adds new bindings for other languages that upstream supports, tidied up manifests
- CI is more robust using new upstream workflows
- Improved queries
- Updated tests, and removing duplicates/excessive tests. There's no need for one test per line of code/tree sitter rule, it just adds unnecessary higher-level nodes when several of these tests can be combined into one test to test them all together
- Rewrote grammar to be more in-tune with the tree-sitter style, including getting rid of useless precedences, rules, hidden or not, inlining where appropriate, supertyping relevant rules, unhiding certain rules/supertypes, etc.
- Improved preproc calls by having content be nested inside these nodes, but this does break the case where it's in the middle of an if statement for example, similar to C. I think the tradeoff is worth it, only ~80 files out of 8k+ files fail because of this, and we get much nicer parse trees in the common case, which, e.g., improves code folding. Another added benefit is we can use preproc rules in spots where only that rule itself is valid, and not everything. e.g. a preproc containing expressions is valid inside expressions, but this won't be done at the top level.
- Fuzz the scanner for obvious reasons

I will update the tokens for publishing the new version later today (needs pypi).

- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/189
- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/283 (todo - quotes)
- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/310
- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/319
- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/326
- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/331
- Closes https://github.com/tree-sitter/tree-sitter-c-sharp/issues/332


tamasvajk (commented):
Thank you for this fix/improvement/contribution. This is going to be awesome. 

I find this PR rather difficult to review. I think the first commit could also contain the test output changes. That would make it easier to see if anything has been broken.

I think there are a couple of bugs that are (re)introduced with this refactoring.
- some of the contextual keyword don’t seem to work any longer, for example parameter `p` in `void m(scoped p) { }` doesn’t have a type. Probably `scoped` became a modifier.
- `var o = (Int32)(1);` became an invocation with a parenthesised identifier, instead of a cast. https://github.com/tree-sitter/tree-sitter-c-sharp/issues/175
- `var greetings = new (string head, string tail)[3];` is not an array creation expression in the equals clause. https://github.com/tree-sitter/tree-sitter-c-sharp/issues/166

There are a lot of changes that I don't understand. Can you explain what's the benefit of changing these:
- some field names have been removed or changed.
- `null_literal` is gone, so `throw null` is now a `throw_expression` without any child.  
- `equals_value_clause` has been removed from the grammar,
- `type_parameter_constraints_clause` has been reworked, previously we had a list of clauses with target and constraint, now we have a flat list of constraints, in the form of `identifier, type_parameter_constraint+, identifier, type_parameter_constraint+`. Isn't this more difficult to handle in for consumers of the output tree?
- `with_initializer_expression` is gone, which means `with_expression` contains a flat list of children. This way the tree seems to be more difficult to process.

amaanq:
Wow, those two dynamic precedence issues caught me off guard. I noticed some places where prec.dynamic was used unnecessarily, but I didn't think those two were. That was really interesting and after a little digging, it makes sense. This is probably one of few grammars that is very sensitive to dynamic precedence changes. Thanks for the context, I've added them back

Others:

- A lot of the fields were unnecessary, however, if you think some of the ones i removed were necessary, let me know and I can add them back
- I added this back, it's generally preferable to not assign a node to a single literal since it just "absorbs" that literal and you can't directly query for the string literal then, but for consistency with all the other literals it makes sense
- I don't see the point in this, unless it resolved ambiguities/reduced state count but I don't think it did
- I remember this *initially* reduced the state count a lot when I was starting my rewrite, but now I noticed it doesn't, so yeah I can just have it as it was before
- Added back, a field called `initializer` could work though if we just want to tag/distinguish the identifier though

grammar.js
```diff
@@ -32,237 +43,304 @@ const stringEncoding = /(u|U)8/;
...
-    _type_declaration: $ => choice(
-      $.class_declaration,
-      $.struct_declaration,
-      $.interface_declaration,
-      $.enum_declaration,
-      $.delegate_declaration,
-      $.record_declaration,
-      $.record_struct_declaration
+        seq(
+          repeat(alias($.statement, $.global_statement)),
```
hvitved, line 148: This change means that `node-types.json` does not provide any information about what `global_statement` can be. Can we revert it back to `repeat($.global_statement)` with `global_statement: $ => $.statement`?

amaanq: I could, but what's the benefit? Curious

Cargo.toml
```diff
@@ -1,36 +1,27 @@
...
+version = "0.21.0"
 authors = [
-    "Damien Guard <damieng@gmail.com>",
-    "Max Brunsfeld <maxbrunsfeld@gmail.com>",
-    "Martin Midtgaard <martin.midtgaard@gmail.com>",
-    "Sjoerd Langkemper <sjoerd-github@linuxonly.nl>",
-    "Patrick Thomson <patrickt@github.com>",
-    "Noelle Caldwell <noelle.caldwell@microsoft.com>",
-    "Douglas Creager <dcreager@dcreager.net>",
+  "Max Brunsfeld <maxbrunsfeld@gmail.com>",
```
damieng, line 6: Why are we wiping some authors out here?

amaanq:
Well I wasn't aware I had to discuss making large improvements beforehand, nor have I ever done so in any other upstream grammar I maintain.

The moving of tests to a different dir is *necessary*, top-level corpus dirs are unsupported upstream now and must be in `test/corpus`.

The exclusion list is either a parse error where it genuinely was an error (beforehand as well) or it has funky usage of preproc ifs, e.g. in the middle of an if/else statement. I have (imo) improved them such that their contents are children of the `preproc_if` node, much like how C does it, however, this requires that the contents inside be somewhat correctly formed (e.g. a regular statement/expression, wherever the preproc_if is *applicable*). I think this is a reasonable tradeoff for much more navigable trees.

grammar.js
hvitved: We have [a generic Treesitter-based CodeQL extractor framework](https://github.com/github/codeql/blob/main/shared/tree-sitter-extractor/src/generator/mod.rs), which, given a Treesitter grammar (`node-types.json`), produces a matching CodeQL database schema. For example, given `node-types.json` [for Ruby](https://github.com/tree-sitter/tree-sitter-ruby/blob/master/src/node-types.json), we produce [a matching DB schema](https://github.com/github/codeql/blob/main/ruby/ql/lib/ruby.dbscheme#L105). When I point it at the `node-types.json` file in this PR, it currently fails because it doesn't know the type of `global_statement`, and hence it is not able to generate a type union that represent all the possible nodes under a `compilation_unit`. In comparison, the [Ruby `program` rule](https://github.com/tree-sitter/tree-sitter-ruby/blob/master/grammar.js#L108) gives rise to [this](https://github.com/github/codeql/blob/main/ruby/ql/lib/ruby.dbscheme#L997) union type.

Cargo.toml
amaanq: I consider the Cargo.toml authors list to be those that are relevant to the **Rust** crate and core grammar, and can be reached out to for any questions, comments, or concerns. Some of those authors either a) have no direct relation to the rust crate, b) are potentially unreachable/inactive.

grammar.js
hvitved: I guess one way we can work around this is basically to say then that a `compilation_unit` can contain _any_ type of AST node as a child.

amaanq: Gotcha, thanks. I'm okay with adding the rule back, though it would be a smidge better without it (one less reduction)

hvitved: In that case I think it would be nice to add it back in, thanks.

amaanq:
@damieng @tamasvajk the diff for tests should be much more readable now

test/corpus/literals.txt
```diff
@@ -128,71 +128,63 @@ var x = $"{{";
...
+          (integer_literal)))))
   (global_statement
     (local_declaration_statement
       (modifier)
       (variable_declaration
         type: (predefined_type)
         (variable_declarator
           name: (identifier)
-          (equals_value_clause
-            (integer_literal))))))
```
maxbrunsfeld, line 195: 👍 to getting rid of `equals_value_clause`.

test/corpus/classes.txt
```diff
@@ -83,119 +89,114 @@ file class A {}
   (class_declaration
     (modifier)
     name: (identifier)
-    body: (declaration_list))
+    (declaration_list))
```
maxbrunsfeld, line 92: IMO the `body` field was fine, but I don't know if there was some drawback to it. Curious if you were seeing significant code size increases due to that field.

queries/highlights.scm
```diff
@@ -179,77 +195,19 @@
...
 
 ;; Type constraints
-(type_parameter_constraints_clause (identifier) @property.definition)
-
-;; Switch
-(switch_statement (identifier) @variable)
-(switch_expression (identifier) @variable)
 
-;; Lock statement
-(lock_statement (identifier) @variable)
```
maxbrunsfeld, line 252: 👍 to removing all of these specific patterns that match `@variable`

maxbrunsfeld:
Hope you're able to reduce the state count. In general, a great way to do that is avoid long sequences with many variations.

In other words, turn this:

```js
seq(
  optional('foo'),
  optional('bar'),
  optional('baz'),
  optional('quux'),
  // ...
)
```

into this:

```js
seq(
  repeat(choice('foo', 'bar', 'baz', 'quux')),
  // ...
)
```

It seems like there may be opportunities to do some of that in this grammar - there may be unnecessarily-specific sequences that could be modeled more generically.

test/corpus/classes.txt
damieng: There's a few of these removed in places that @tamasvajk added - are they used by GH Semantic?

test/corpus/attributes.txt
```diff
@@ -8,15 +8,15 @@ Global attributes
...
     (attribute
       name: (identifier)))
-  (global_attribute_list
+  (global_attribute
     (attribute
       name: (identifier))
     (attribute
       name: (identifier)
-      (attribute_argument_list))))
+      (attribute_arguments))))
```
damieng, line 19: These changes are going to be breaking and they're going to take us away from the Roslyn grammar definitions.

e.g. attribute_argument_list being wiped out and attribute_argument becoming attribute_arguments.

Originally I aligned with them because it made it easier to write the grammar but it's also been really helpful as the language evolves to see which rules they changed for the new features and how ours changes too. It's also provided a great way to know we're parsing the code correctly as we can match the definitive parser.

If moving away from that is desirable because it makes the grammar a lot smaller or enables some new tree-sitter functionality it might be worth the trade-off but if not I'd prefer to retain the alignment and compatibility.

test/corpus/contextual-keywords.txt
```diff
@@ -188,9 +184,9 @@ class scoped { }
       name: (identifier)
       parameters: (parameter_list
         (parameter
-          (parameter_modifier)
-          (parameter_modifier)
-          type: (predefined_type)
+          type: (scoped_type
```
damieng, line 187: This looks very off? They've gone from parameter modifiers to type modifiers?

test/corpus/expressions.txt
```diff
@@ -19,63 +19,54 @@ a = a!;
     (expression_statement
       (assignment_expression
         left: (identifier)
-        (assignment_operator)
```
damieng, line 22: This one I'm fine with - it aligns better with Roslyn.

test/corpus/expressions.txt
```diff
@@ -405,35 +366,29 @@ var x = new A
...
         right: (integer_literal))))
   (global_statement
     (expression_statement
-      (assignment_expression
-        left: (prefix_unary_expression
-          (identifier))
-        (assignment_operator)
-        right: (integer_literal)))))
+      (prefix_unary_expression
```
damieng, line 388: Subtle change here with the assignment now being a prefix_unary_expression rather than the left term.

test/corpus/expressions.txt
```diff
@@ -1167,63 +1088,53 @@ void m() {
...
-                (anonymous_method_expression
-                  (modifier)
-                  parameters: (parameter_list
-                    (parameter
-                      type: (predefined_type)
-                      name: (identifier)))
-                  (block
-                    (return_statement
-                      (identifier))))))))
+              (anonymous_method_expression
```
damieng, line 1091: Why is it we're losing "static" as a (modifier) here but we're keeping static as a modifier on the lambda equivalent on line 1262?

test/corpus/enums.txt
```diff
@@ -9,7 +9,7 @@ enum A: byte { One, Two = 2, Three = 0x03 }
 (compilation_unit
   (enum_declaration
     name: (identifier)
-    bases: (base_list
+    (base_clause
```
damieng, line 12: base_list in Roslyn - not sure why we'd change it without a reason.

test/corpus/interfaces.txt
```diff
@@ -129,34 +128,31 @@ public interface IGetNext<T> where T : IGetNext<T>
...
         body: (declaration_list
-          (event_field_declaration
-            (variable_declaration
-              type: (generic_name
-                name: (identifier)
-                type_arguments: (type_argument_list
-                  (identifier)))
-              (variable_declarator
-                name: (identifier))))
+          (event_declaration
```
damieng, line 150: Again this has gone from a definitive Roslyn-defined set to what appears to be an arbitrary choice/level of abstraction.

Here's the section from syntax analyzer:

image

test/corpus/preprocessor.txt
```diff
@@ -153,36 +149,36 @@ class Of1879 {
...
-            (preproc_integer_literal)
-            (preproc_integer_literal)
-            (preproc_integer_literal)
-            (preproc_string_literal))
-          (line_directive
-            (preproc_integer_literal)
-            (preproc_string_literal))
-          (line_directive)
-          (line_directive))))))
+          (preproc_call
```
damieng, line 158: This feels like a definite step backwards in terms of usefulness.

amaanq: do you mean the granular parsing of the line preprocessor directive? this is more how C handles it, but C# might be different because it has a fixed set of directives. I could add back the granularity, but I initially thought it's probably better to handle it like C

test/corpus/expressions.txt
amaanq: Gotcha, but what about static/unsafe in using_directive? I also (personally) think it's ok to hide modifiers, and just expose the string literals in the tree instead, but I don't have much of a preference.

test/corpus/expressions.txt
amaanq: good catch - that original splitting of _pointer_indirection_expression was very costly due to the conflict - but if we only allow lvalue_expression's there then it's much better

test/corpus/interfaces.txt
amaanq: no *this* one doesn't improve it all that much, to me I just thought that `event_declaration` satisfies this enough. this is innocuous enough to add back, but im curious why do we try to strictly follow the roslyn definition? I feel like that can cause some pain points in certain spots, but in this case it just adds a conflict that is sorta unnecessary, though state count only goes up a *smidge*

damieng: Why would we not want to use the definitive name for the grammar parts rather than whoever adds a feature choose arbitrary names?

amaanq:
Down to 9700/3200 after reworking top level stuff + preprocs 😁

test/corpus/literals.txt
```diff
@@ -266,528 +250,518 @@ var x = $"{{";
...
-                (identifier))
-              (interpolated_string_text)))))))
+          (interpolated_string_expression
+            (interpolation_start)
+            (string_content)
+            (interpolation
+              (interpolation_open_brace)
+              (identifier)
+              (interpolation_close_brace))
+            (string_content))))))
```
tamasvajk, line 485: Why do we include the start, quote and brace tokens in the tree?

tamasvajk (commented):
Thanks for rebasing the PR, it's a lot easier to check the changes in the test output. I've added some comments/questions.

test/corpus/classes.txt
tamasvajk: I think the fields are mostly used by GH CodeQL. @hvitved knows best how much we need them.

hvitved: Having field names in general certainly makes the life easier for Treesitter based QL extractors. Field names are used to generate methods in CodeQL, e.g. [this field](https://github.com/tree-sitter/tree-sitter-ruby/blob/9d86f3761bb30e8dcc81e754b81d3ce91848477e/grammar.js#L196) in the Ruby grammar gives rise to [this QL predicate](https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/ast/internal/TreeSitter.qll#L1443), which is nicer to use than something like `getChild(0)`.

test/corpus/literals.txt
amaanq: they can be variable in size, i could hide them but this then becomes unqueryable, or i could alias them to a literal, e.g. alias open_brace to '{', but that wouldn't always be correct since it could be multiple

tamasvajk: Makes sense. I'd keep it as it is. It just seemed unusual to see them in the tree.

grammar.js
```diff
@@ -281,148 +362,140 @@ module.exports = grammar({
       // 'ref',     // `ref` as a modifier can only be used on struct declarations. Other than that it's a ref type or a ref parameter in a declaration.
       // 'scoped',  // `scoped` is either part of a scoped type or a scoped parameter. Both of which are handled outside of `modifier`.
       'sealed',
-      'static',
+      'static', // hmm?
```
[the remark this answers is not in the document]
amaanq: I think I was just debugging something and left that as a note to self, oops

test/corpus/interfaces.txt
```diff
@@ -170,7 +168,7 @@ public interface IGetNext<T> where T : IGetNext<T>
     name: (identifier)
     body: (declaration_list
       (method_declaration
-        type: (predefined_type)
+        returns: (predefined_type)
```
damieng, line 171: @hvitved is this shift from type to returns going to cause upstream issues for semantic?

test/corpus/type-methods.txt
```diff
@@ -247,21 +260,20 @@ ref struct S {
...
     body: (declaration_list
       (method_declaration
-        type: (predefined_type)
+        returns: (predefined_type)
         name: (identifier)
         parameters: (parameter_list
           (parameter
-            (parameter_modifier)
-            (parameter_modifier)
+            (modifier)
```
damieng, line 270: I'm okay with the parameter_modifier > modifier change

test/corpus/attributes.txt
```diff
@@ -79,10 +79,9 @@ class Zzz {
       (attribute
         name: (identifier)
         (attribute_argument_list
-          (attribute_argument
-            (member_access_expression
-              expression: (identifier)
-              name: (identifier))))))
+          (member_access_expression
```
damieng, line 82: Maybe put back attribute_argument back between these two for consistency/non-breaking.

test/corpus/preprocessor.txt
```diff
@@ -13,39 +13,32 @@ If, elif and else directives
 --------------------------------------------------------------------------------
 
 (compilation_unit
-  (if_directive
-    (identifier))
-  (global_statement
+  (preproc_if
```
damieng, line 16: Let's see how we get on with these if > preproc changes - they're not part of Roslyn and probably less used by third party tooling relying on this library.

test/corpus/type-fields.txt
```diff
@@ -125,52 +124,40 @@ struct A {
...
           type: (function_pointer_type
             (function_pointer_parameter
               type: (function_pointer_type
-                (function_pointer_calling_convention
-                  (function_pointer_unmanaged_calling_convention_list
-                    (function_pointer_unmanaged_calling_convention
-                      (identifier))
-                    (function_pointer_unmanaged_calling_convention
-                      (identifier))))
+                (calling_convention
```
damieng, line 151: The function pointer calling convention has the old term in the grammar but it's definitely an less used construct so if this shared calling_convention results in an easier/shorter grammar I'm 👍🏻

test/corpus/interfaces.txt
hvitved: I don't work on Semantic, but rather CodeQL. We don't currently have an official tree-sitter based C# CodeQL extractor, so nothing breaks for us (as long as we can still correctly parse what we could earlier).

damieng:
This is shaping up great! If we can get attribute_argument back in then I'm good on the Roslyn alignment/back compat from the naming side.

I think we just need to check/adjust the changes and removals of fields to make sure Semantic doesn't break and we should be good to go.

dcreager:
> I think we just need to check/adjust the changes and removals of fields to make sure Semantic doesn't break and we should be good to go.

Thanks for checking in, :+1: from our side!  We're pinned to the current release so we won't silently upgrade, and we're using the syntax highlighting queries directly from this repo, which I see are updated as part of this PR.  We have augmented the tagging queries with the ability to build up scoped names, which we'll have to update as part of bumping to the new version containing these changes.  But the changes to `tags.scm` in this PR look manageable, so I don't consider that a blocker.

damieng (approved):

amaanq:
thanks for the review/feedback @maxbrunsfeld @damieng @tamasvajk @hvitved!

hvitved:
> * Improved preproc calls by having content be nested inside these nodes, but this does break the case where it's in the middle of an if statement for example, similar to C. I think the tradeoff is worth it, only ~80 files out of 8k+ files fail because of this, and we get much nicer parse trees in the common case, which, e.g., improves code folding. Another added benefit is we can use preproc rules in spots where only that rule itself is valid, and not everything. e.g. a preproc containing expressions is valid inside expressions, but this won't be done at the top level.

Is there any way, using `extras` as before this PR, to still tolerate `#if`s that happen inside expressions or statements? For example, [this file ](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs) appears to have more parse errors now than before.

PR #334, opened 2024-05-01
Cleanup

amaanq:
only adds binding files and moves tests, meant to make the diff in #333 more readable


tamasvajk:
Do you happen to know why the windows test is failing?

amaanq:
outdated prebuildify + old workflow combination, #333 runs fine

PR #335, opened 2024-05-06
Include literal content in the parse tree

hvitved:
https://github.com/tree-sitter/tree-sitter-c-sharp/pull/333 follow-up. While it is [in principle possible to derive that character literal content](https://github.com/github/codeql/pull/16433), it's probably better to revert this change to the grammar.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1734,9 +1734,11 @@ module.exports = grammar({
 
     character_literal: $ => seq(
       '\'',
-      choice(token.immediate(/[^'\\]/), $.escape_sequence),
+      choice($.character_literal_content, $.escape_sequence),
       '\'',
     ),
+
+    character_literal_content: $ => token.immediate(/[^'\\]/),
 
     integer_literal: _ => token(seq(
       choice(
--- a/package-lock.json
+++ b/package-lock.json
@@ -1374,9 +1374,9 @@
       }
     },
     "node_modules/tree-sitter-cli": {
-      "version": "0.22.5",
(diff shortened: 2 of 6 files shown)


grammar.js
```diff
@@ -1734,10 +1734,12 @@ module.exports = grammar({
 
     character_literal: $ => seq(
       '\'',
-      choice(token.immediate(/[^'\\]/), $.escape_sequence),
+      choice($.character_literal_content, $.escape_sequence),
       '\'',
     ),
 
+    character_literal_content: $ => token.immediate(/[^'\\]/),
+
```
amaanq, line 1742:
```suggestion
      choice(
        alias(token.immediate(/[^'\\]/), $.character_literal_content),
        $.escape_sequence
      ),
      '\'',
    ),
```

hvitved: Why is this version preferable? And won't it have the same issue that we discussed [here](https://github.com/tree-sitter/tree-sitter-c-sharp/pull/333#discussion_r1586422465)?

amaanq: ah, right. it's just a little less noisy (since it's not reused anywhere else) and is sorta often done in other grammars. it's not a big difference imo, so this is fine

hvitved:
> also please regenerate with the _latest_ cli (0.22.6) (you can install it with `npm install -g tree-sitter-cli` or `cargo install tree-sitter-cli`)

Done. Thanks for the quick review.

hvitved:
> would it benefit you if I tagged a new release for this?

No need.

PR #336, opened 2024-05-07
Parse arguments in base lists

hvitved:
See https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors#initialize-base-class.


amaanq:
nice!!

PR #354, opened 2024-11-01
Work-around Tree-Sitter regex inability to match certain unicode chars.

gundermanc:
I noticed that Tree-Sitter-C-Sharp is unable to match certain types of Unicode characters, like those prevalent in Chinese. In my testing, the C++ grammar doesn't seem to have this problem.

This PR ports the [identifier regex used by C++](https://github.com/tree-sitter/tree-sitter-cpp/blob/master/src/grammar.json#L10899) over to C#, dropping just the part of it that matches `$`, enabling correct parsing of identifiers that contain Chinese characters.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1827,7 +1827,7 @@ module.exports = grammar({
 
     boolean_literal: _ => choice('true', 'false'),
 
-    _identifier_token: _ => token(seq(optional('@'), /[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Pc}\p{Cf}\p{Mn}\p{Mc}]*/)),
+    _identifier_token: _ => token(seq(optional('@'), /(\p{XID_Start}|_|\\u[0-9A-Fa-f]{4}|\\U[0-9A-Fa-f]{8})(\p{XID_Continue}|\\u[0-9A-Fa-f]{4}|\\U[0-9A-Fa-f]{8})*/)),
     identifier: $ => choice(
       $._identifier_token,
       $._reserved_identifier,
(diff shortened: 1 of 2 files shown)


gundermanc:
I had to touch a few other places to get the linter to pass.

amaanq:
> I had to touch a few other places to get the linter to pass.

No worries, that's on me for not updating it when I added the new eslint config

As for the actual change, I'm not 100% sure this is the correct pattern for valid C# identifiers, but it looks fine to me. I wonder if regenerating with the latest tree-sitter master would actually fix your issue with Chinese unicode characters as identifiers, since that was a bug fixed recently with our unicode tables 🤔

gundermanc:
> I'm not 100% sure this is the correct pattern for valid C# identifier

My understanding is that identifiers have an optional @ at the start (not reflected in the grammar before, but I added this revision), followed by one letter, and 0 or more alpha numeric characters. I'm not sure what this means in Unicode terms, but I've tried some examples of Chinese characters, and they seem to match consistent with how they compile with the .NET compiler and VS intellisense.

> do you mind trying that out

It still fails with tip of main Tree-Sitter.

> pasting the valid code that fails to parse so I can try it out?

Yep, this compiles in VS in C#.

```csharp
            class Class狧麱狵錋狾龍龪啊阿埃挨哎唉0u㐀㐁㐂㐃㐄㐅㐆㐇6ⅶ０ǒoU1U2U38丂丄丅丆丏丒丟
            {
                void Test狧麱狵錋狾龍龪啊阿埃挨哎唉0u㐀㐁㐂㐃㐄㐅㐆㐇6ⅶ０ǒoU1U2U38丂丄丅丆丏丒丟()
                {
                    Console.WriteLine("Hello World!");
                }
            }
```

gundermanc:
Hey @amaanq, do you mind taking another look when you have a chance?

Thank you again for your time!

amaanq:
thanks, sorry was on a trip

PR #355, opened 2024-11-05
Pattern matching improvements

vanessa-rodrigues:
- Allows single sub pattern in positional pattern clauses.
- Allows variable designation in list patterns.
- Allows name_colon in sub pattern.

Reference:
https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -46,6 +46,7 @@ module.exports = grammar({
   conflicts: $ => [
     [$._simple_name, $.generic_name],
     [$._simple_name, $.type_parameter],
+    [$._simple_name, $.subpattern],
 
     [$.tuple_element, $.type_pattern],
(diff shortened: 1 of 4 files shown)


vanessa-rodrigues:
Parsing examples/nunit/src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs seems to fail even before my changes.

amaanq:
> Parsing examples/nunit/src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs seems to fail even before my changes.

Yeah that's fine, looks like a file missing from the excluded list in CI. Do you mind rebasing to resolve the conflicts?

damieng (approved):

PR #360, opened 2024-12-05
Re-expose HIGHLIGHTS_QUERY.

patrickt:
It looks like this was exposed in v0.21, but regressed with the release of 0.23.


hendrikvanantwerpen (approved):
🫠

PR #364, opened 2025-01-06
Support bodyless class/struct/record/enum/interface

### LINKED ISSUE #339: Types without curly braces are lacking support
Type declarations such as:

```cs
class MyClass1;
class MyClass2();
interface IMyInterface1;
interface IMyInterface2();
enum MyEnum1;
enum MyEnum2();
```

show up as errors in the syntax tree. The README says that the project supports C# up to v13, so these declarations should be supported.

damieng:
Fixes #339

Note: 

- Original issue has `enum()` and `interface()` in list of samples - those are not supported in C# as far as I know (latest VS does not support that syntax either)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,8 @@ dist/
 *.tar.gz
 *.tgz
 *.zip
+
+# Build artifacts
+parser.exp
+parser.lib
+bindings/*
--- a/grammar.js
+++ b/grammar.js
@@ -265,7 +265,10 @@ module.exports = grammar({
 
     class_declaration: $ => seq(
       $._class_declaration_initializer,
-      $._optional_semi,
+      choice(
+        seq(field('body', $.declaration_list), $._optional_semi),
+        ';'
+      )
     ),
 
     _class_declaration_initializer: $ => seq(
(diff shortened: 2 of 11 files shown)


maxbrunsfeld:
Nice!

amaanq:
Hey @damieng, I think your CLI is a bit outdated, as you made some changes to package.json, .gitignore, and other files that should not have been made, do you mind updating & regenerating?

damieng:
I can revert the package.json changes no problem.

I'm using `tree-sitter 0.24.6 (21a517c423010811147b0b1aa1e7aedc39ce91aa)` and it definitely generates a `parser.exp` and `parser.lib` file here (on Windows) and git is wanting to track those without the .gitignore additions.

amaanq:
🤔 interesting, do you want to submit a pr that denies all `*.exp` and `*.lib` files [upstream](https://github.com/tree-sitter/tree-sitter/blob/ba19fe31bed96a28aa5a1b9253a716e494ed7a9b/cli/src/templates/gitignore#L27)? It'd be better to do it here, then let the CLI update this file when the next release is out

PR #365, opened 2025-01-06
build: revert changes triggered by old tree-sitter-cli

damieng:
Reverts undesirable changes from https://github.com/tree-sitter/tree-sitter-c-sharp/commit/d13ccdd7fb36f94ce89b8a52ae4bd2f2f0c9f096

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,7 +39,3 @@ dist/
 *.tgz
 *.zip
 
-# Build artifacts
-parser.exp
-parser.lib
-bindings/*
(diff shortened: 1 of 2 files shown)


amaanq:
can you also regenerate? the src/ files should have some slight changes I'd think

damieng:
Not seeing changes.

I think the package.json and bindings folder may have been left over from a previous run before I updated tree-sitter-cli but the generated files were up to date.

changes

amaanq:
oh yeah, my bad

PR #367, opened 2025-02-06
feat: preprocess entire regions

wizardlink:
Hello! First time messing with `tree-sitter`, so apologies if my changes are too drastic.

I've noticed recently while using [nvim-treesitter] that there are no folds for `#region` directives, unlike [tree-sitter-c](https://github.com/tree-sitter/tree-sitter-c/blob/master/grammar.js#L1409).

This is due to the fact that the `tree-sitter-c-sharp` parser has no rule that encapsulates the content of the directive alongside itself, only the directives themselves through `preproc_region` and `preproc_endregion`.

If I am meant to implement this rule in [nvim-treesitter], I apologise in advance! Like I mentioned, first time messing with this.

[nvim-treesitter]: https://github.com/nvim-treesitter/nvim-treesitter

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -116,8 +116,7 @@ module.exports = grammar({
   extras: $ => [
     /[\s\u00A0\uFEFF\u3000]+/,
     $.comment,
-    $.preproc_region,
-    $.preproc_endregion,
+    $.preproc_full_region,
     $.preproc_line,
     $.preproc_pragma,
     $.preproc_nullable,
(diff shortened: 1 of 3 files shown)


damieng:
Hi Wizardink, thanks for the PR.

I'm not entirely convinced we want to do this. It makes all the AST inside the region child nodes of the region.

As the regions are pre-processor directives and not part of the C# language this isn't correct and is quite at-odds with how the C# language is normally parsed such as by Roslyn (which treats the opening as leading trivia to the statement that proceeds it and end trivia for the statement it follows.

Open to more discussion but given this would likely be a breaking change for people using the grammar for things other than editors - regions could now split code into disparate chunks - we need to be sure this it the way to go.

damieng:
I'm not really familiar with the higher levels beyond the parsing but when discussing this with @amaanq he suggested:

> it might be better to "simulate" the region with queries, by just querying for a region + endregion and capturing anything in between them

PR #408, opened 2026-02-09
chore: add schema link to `tree-sitter.json`

jfly:


jfly:
Asked upstream: https://github.com/tree-sitter/tree-sitter/pull/5315

damieng:
@tris203 Is there any reason not to merge this then? Or should we have the discussion first?

tris203:
I think that it might be breaking for downstream

jfly:
I don't see how this diff could break anything: it just adds a link to the schema for the file, AFAICT, nothing will enforce it.

tris203:
> I don't see how this diff could break anything: it just adds a link to the schema for the file, AFAICT, nothing will enforce it. 

My bad. I thought the name was still changed to remove hyphens. 
Schema only is LGTM

Issue #401, opened 2026-01-16
Add Support for C# 12 Collection Expressions


gustavoaca1997:
## Summary

The tree-sitter C# grammar does not currently support **C# 12 Collection Expressions** (`[]` syntax), introduced in .NET 8 / C# 12. This causes parse errors when processing modern C# code that uses this syntax.

## Description

Collection expressions are a concise syntax for initializing collections in C# 12. The parser currently fails with an error like `(identifier): '<empty>'` when encountering these expressions, as it expects an identifier or expression where `[]` appears.

**Microsoft Documentation**: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions

## Reproduction

The following C# code patterns fail to parse:

### 1. Empty collection literal
```csharp
var empty = [];
List<int> list = [];
```

### 2. Method return values
```csharp
public List<string> GetItems()
{
    return [];
}
```

### 3. Method arguments
```csharp
SomeMethod([]);
mock.Setup(x => x.GetData()).ReturnsAsync([]);
```

### 4. Variable initialization
```csharp
List<OssMatch> ossMatches = [];
```

### 5. Null-coalescing fallback
```csharp
var items = knownItems?.Select(x => x.Value) ?? [];
```

### 6. Collection with elements
```csharp
var numbers = [1, 2, 3, 4, 5];
```

### 7. Spread operator
```csharp
var combined = [..first, ..second];
var mixed = [1, 2, ..others, 3, 4];
```

## Expected Behavior

All the above code examples should parse successfully without errors.

## Actual Behavior

Parse errors occur at the `[]` token with messages like:
```
Parse errors with language 'csharp' at 'File.cs'(1 error(s)):
  [1] Line X, Col Y (identifier): '<empty>'
```

## Real-World Examples

These patterns appear frequently in modern C# codebases. Here are some specific cases encountered:

**Mock setups (Moq)**:
```csharp
_mockService.Setup(p => p.ProcessAsync(It.IsAny<Request>()))
    .ReturnsAsync([]);  // Parse error at Col 31
```

**Early returns**:
```csharp
if (IsEmpty(result))
{
    return [];  // Parse error
}
```

**Constructor arguments**:
```csharp
var set = new HashSet<string>(items ?? [], StringComparer.OrdinalIgnoreCase);
```

## Proposed Solution

_I can try to implement this._

Add support for collection expressions to the grammar by introducing new rules:

```javascript
[...]

PR #409, opened 2026-02-11
feat: add C# 12 collection expression support

slang25:
## Summary

- Implements grammar rules for collection expressions (`[e1, e2, ..e3]`) and spread elements (`..`)
- Resolves syntax conflicts with list patterns and range expressions through precedence rules and GLR declarations
- Adds `element_binding_expression` handling to initializer expressions to preserve dictionary/indexer initialization syntax
- All 164 existing tests pass, plus 9 new collection expression test cases

## Test Coverage

Collection expressions tested for: empty literals, simple elements, spread operators, trailing commas, nested collections, method calls in spreads, use as arguments, and conditional spreads with parentheses.


guillaume86:
I independently worked on the same feature in #414 and validated this PR against the pathological case from [my issue comment](https://github.com/tree-sitter/tree-sitter-c-sharp/issues/401#issuecomment-4229726773): `(int[])[Constants.Region.US, Constants.Region.EU, ...]` inside a `foreach` was causing greedy error recovery that swallowed subsequent method declarations.

Confirmed this PR fully fixes it — all methods parse correctly with zero ERROR nodes. Ternary + collection (`condition ? [x] : [y]`) also works.

The `_initializer_index_assignment` approach to preserve `element_binding_expression` in initializer contexts is cleaner than what I had. 👍

Note: the CI parse-examples step will likely fail due to pre-existing issues — PowerToys and NUnit repos have been updated since the last master CI run, introducing files at new paths that aren't in the exclude list yet. See the exclude list update in my (now-closed) #414 for the specific files that need adding.

PR #410, opened 2026-02-16
Fix C# 13 params non-array types

damieng:
Allow C# 13 params that use collection types instead of arrays https://csharp-evolution.com/13.0/params-collections

Also fixes invalid esc sequences, dupe conversion operations and tightens up arg attributes.

Claude Code was used in the development of this PR.


grammar.js
```diff
@@ -481,11 +479,9 @@ export default grammar({
         'implicit',
         'explicit',
       ),
-      repeat1(choice(
-        $.explicit_interface_specifier,
-        'operator',
-        'checked',
-      )),
```
tris203, line 488: This was something i changed as a perf optimisation last year. 
I would prefer not to change it back

tris203 (approved):

grammar.js
damieng: I'll revert this back and put a note in.

PR #414, opened 2026-04-13
Add C# 12 collection expressions and spread element support

### LINKED ISSUE #401: Add Support for C# 12 Collection Expressions

guillaume86:
## Summary

- Adds `collection_expression` rule for C# 12 collection expression syntax (`[1, 2, 3]`, `[]`, `[..a, ..b]`)
- Adds `spread_element` rule for the `..expr` spread syntax within collection expressions
- Replaces standalone `element_binding_expression` in `lvalue_expression` with `collection_expression`, following the C# spec disambiguation approach
- Adds GLR conflict + `prec.dynamic(1)` on `list_pattern` so `x is [...]` still resolves to list pattern in pattern contexts

Fixes #401 — including the [severe error recovery case](https://github.com/tree-sitter/tree-sitter-c-sharp/issues/401#issuecomment-4229726773) where `(int[])[Constants.Region.US, ...]` caused subsequent method declarations to be swallowed by ERROR nodes.

## Test plan

- [x] All 158 existing tests pass (no regressions)
- [x] 11 new tests added covering: empty, elements, spread, mixed, method arguments, return values, cast+collection, null-coalescing, nested, range-inside-collection, and the exact reproduction case from the issue comment
- [x] Parser state count increase is minimal (+1.4%, 7986 → 8099 states)

### DIFF
--- a/test/corpus/expressions.txt
+++ b/test/corpus/expressions.txt
@@ -373,9 +373,8 @@ var x = new A
                 left: (identifier)
                 right: (integer_literal))
               (assignment_expression
-                left: (element_binding_expression
-                  (argument
-                    (identifier)))
+                left: (collection_expression
+                  (identifier))
                 right: (integer_literal))))))))
   (global_statement
     (expression_statement
@@ -2387,10 +2386,9 @@ var x = new Dictionary<string,int> { ["a"] = 65 };
                 (predefined_type)))
             initializer: (initializer_expression
               (assignment_expression
-                left: (element_binding_expression
-                  (argument
-                    (string_literal
-                      (string_literal_content))))
+                left: (collection_expression
+                  (string_literal
+                    (string_literal_content)))
                 right: (integer_literal)))))))))
 
 ================================================================================
@@ -3142,9 +3140,8 @@ var x = dict?["a"];
           (conditional_access_expression
             condition: (identifier)
             (element_binding_expression
-              (argument
-                (string_literal
-                  (string_literal_content))))))))))
+              (string_literal
+                (string_literal_content)))))))))
 
 ================================================================================
 Conditional access expression with member binding
@@ -3186,8 +3183,7 @@ if ((p as Person[])?[0]._Age != 1) { }
                   type: (identifier)
                   rank: (array_rank_specifier))))
             (element_binding_expression
-              (argument
-                (integer_literal))))
+              (integer_literal)))
           name: (identifier))
         right: (integer_literal))
       consequence: (block))))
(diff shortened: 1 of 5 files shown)


guillaume86:
## CI fix update

The parse-examples CI step was failing on 8-13 files. After investigation, I confirmed these are **pre-existing failures** from the example repos (PowerToys, NUnit) having been updated with newer C# syntax since the last master CI run — they also fail on a clean upstream master clone with the latest example repo contents.

Specific causes:
- PowerToys reorganized their `cmdpal` directory (files moved from `Core/` paths to new paths, no longer matching existing excludes)
- PowerToys and NUnit added code using C# 12 collection expressions and list patterns with slices that the grammar doesn't fully support yet

I've added the affected files to the CI exclude list in a separate commit.

test/corpus/expressions.txt
```diff
@@ -2387,10 +2386,9 @@ var x = new Dictionary<string,int> { ["a"] = 65 };
                 (predefined_type)))
             initializer: (initializer_expression
               (assignment_expression
-                left: (element_binding_expression
-                  (argument
-                    (string_literal
-                      (string_literal_content))))
+                left: (collection_expression
+                  (string_literal
+                    (string_literal_content)))
```
tris203, line 2391: Thanks, could you take a look at #409. This seems to be trying to do the same thing and seems to avoid changing this node

tris203 (commented):
Thanks for following up with the CI

guillaume86:
After reviewing #409, I can confirm it takes a cleaner approach to the same problem — specifically the `_initializer_index_assignment` pattern to preserve `element_binding_expression` in initializer contexts.

I validated #409 against the exact pathological case from my [issue comment](https://github.com/tree-sitter/tree-sitter-c-sharp/issues/401#issuecomment-4229726773): the cast + collection with qualified member access (`(int[])[Constants.Region.US, ...]`) that was swallowing subsequent method declarations. It's fully fixed — all 3 methods parse correctly with zero ERROR nodes.

Also confirmed: ternary + collection (`condition ? [x] : [y]`) works correctly too.

Closing in favor of #409.

test/corpus/expressions.txt
guillaume86: Ah yes it's cleaner, #409 + this CI fix and it's good IMO 👍

PR #422, opened 2026-04-26
feat: add highlighting for unsafe block statement

Strafe153:
So far, the `unsafe` keyword is highlighted when it's used as a member modifier, however the highlighting is missing for an unsafe code block

Before:
image

After:
image


damieng:
Thanks!

PR #3, opened 2017-11-23
Support for nested-usings, interfaces and methods

Xapphire13:

Issue #1, opened 2017-11-22
Can't build on Windows


Xapphire13:
When I run `npm run build` I get the following error:

```
binding.obj : error LNK2001: unresolved external symbol ts_language_c_sharp [D:\github\tree-sitter-c-sharp\build\ts_language_c_sharp_binding.vcxproj]
D:\github\tree-sitter-c-sharp\build\Release\ts_language_c_sharp_binding.node : fatal error LNK1120: 1 unresolved externals [D:\github\tree-sitter-c-sharp\build\ts_language_c_sharp_binding.vcxproj]
```

Running **Windows 10**
NPM **5.4.2**
Node **8.7.0**

Xapphire13:
Ironically, it builds fine on Linux (running on the Windows Linux Subsystem).

Kernel **4.4.0**
Ubuntu *16.04**

Issue #392, opened 2025-09-30
C# 14 support


damieng:
Now that C# version 14 is settling down we need to look at what changes are needed to support new syntax.

- [ ] [Extension members](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#extension-members)
- [ ]  [Null-conditional assignment](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#null-conditional-assignment)
- [ ]  [nameof supports unbound generic types](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#unbound-generic-types-and-nameof)
- [x] [More implicit conversions for Span<T> and ReadOnlySpan<T>](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#implicit-span-conversions)
- [ ]  [Modifiers on simple lambda parameters](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#simple-lambda-parameters-with-modifiers)
- [ ]  [field backed properties](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#the-field-keyword)
- [ ]  [partial events and constructors](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#more-partial-members)
- [ ]  [user-defined compound assignment operators](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#user-defined-compound-assignment)

kasperk81:
c# 15 right around the corner https://devblogs.microsoft.com/dotnet/explore-csharp-15/

PR #402, opened 2026-01-17
feat: collection expression support

gustavoaca1997:
This PR closes #401 

This is my first time contributing to a Tree Sitter grammar, so please let me know if you have any concerns, suggestions, extra test cases, etc.

## Change details
- Add new test cases to expressions.txt.
- Modify the test case `Array with trailing comma`. Roselyn is actually matching `[ y, ]` against `collection_expression`, so I've updated the test to reflect that. I used [SharpLab](https://sharplab.io/#v2:C4LghgzsA0AmIGoA+ABATARgLACgUGYACdQgYUIG9dCbiiUAWQgWQAoBKS62ngNzABOhAJ4BeAAwBubjxoBLAHbAA2gF1CAD0KjCykdEKrpOWbUUr1AL22bjPAL657QA) to verify that.
- Add grammar rules for `collection_element` , `expression_element` , ` spread_element` , ` collection_expression` 

## How the changes were verified
I ran the tests locally, to verify they are passing.

Closes #409

PR #429, opened 2026-05-19
feat: C# 14 language support

AlexLaroche:
## Summary

Adds support for the seven syntactic features introduced in [C# 14](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14):

| # | Feature | Path |
|---|---|---|
| 1 | `nameof` with unbound generic types (`nameof(List<>)`) | test-only — already worked via `unbound_type_name` |
| 2 | Null-conditional assignment (`obj?.x = v`, `obj?[i] = v`) | grammar |
| 3 | Partial events and partial constructors | test-only — already worked via existing `partial` modifier |
| 4 | User-defined compound assignment operators + instance `++`/`--` | grammar |
| 5 | Field-backed properties (`field` contextual keyword + mix of auto/full accessors) | test-only — `field` is an identifier in this position |
| 6 | Extension declarations (`extension(...) { ... }` blocks in static classes) | grammar |
| 7 | Simple-lambda parameters with modifiers (`(ref x) => x`, `(out y) => ...`, `(in z) => ...`, `(text, out result) => ...`) | grammar + external scanner |

Closes part of #392.

The README's status section is updated to reflect C# 14 support. File-based-apps preprocessor directives (`#:property`, `#:package`, `#:sdk`, `#:project`) remain unrecognized and are noted as a known limitation alongside the existing `async`/`var`/`await` identifier-use caveat.

## Commits

Eight atomic commits, linear history, one feature per commit (plus the README). Each commit is independently buildable and testable — `tree-sitter generate && tree-sitter test` passes at every step.

```
38fc039 test: pin nameof support for unbound generic types (C# 14)
cb973b6 test: add partial constructors and partial events corpus (C# 14)
6c5c419 test: pin field-backed properties with `field` contextual keyword (C# 14)
d63b68d feat: support null-conditional assignment (C# 14)
dcb72c6 feat: C# 14 user-defined compound assignment and instance increment operators
d69e097 feat: C# 14 extension declarations
7c7b5f3 feat: C# 14 simple-lambda parameters with modifiers
e16cfe9 docs: update README to reflect C# 14 support
```

## Notes on individual features

**Null-conditional assignment** — adds `conditional_access_expression` to the `lvalue_expression` choice with one new conflict declaration so the existing non-lvalue parse is preserved.

**Compound assignment + instance increment/decrement** — extends the operator token choice list to include `+=`, `-=`, `*=`, `/=`, `%=`, `^=`, `|=`, `&=`, `<<=`, `>>=`, `>>>=`. The existing `optional('checked')` keyword covers `checked +=` and `checked ++`. The proposal specifies these as instance members returning `void` with one parameter (compound) or zero (instance increment) — `_function_body` already permits empty `()` parameter lists.

**Extension declarations** — new `extension_declaration` node added to the `declaration` choice. `extension` is a contextual keyword recognized via `prec.dynamic`; one new conflict (`[receiver_parameter, scoped_type, _reserved_identifier]`) keeps `scoped` interpretations alive inside the receiver until later tokens disambiguate.

**Simple-lambda parameters with modifiers** — the only feature requiring an external scanner. At `(`-followed-by-a-parameter-modifier, the LR table has already resolved the action to `parenthesized_expression(ref_expression(...))` or `tuple_expression(argument, ...)` due to `lambda_expression`'s `prec(-1)`. No combination of conflict declarations or precedence rules keeps the lambda interpretation alive without breaking existing parse-tree shapes (every grammar-only attempt regressed at least one of *Lambda expression with ref modifier*, *Class method with single parameter*, or *Scoped contextual keyword*). Roslyn and every production C# parser solve this with unbounded backward-tracking lookahead — this PR mirrors that approach with a single external token `_lambda_paren_open` emitted by the scanner only when forward-scanning confirms a closing `)=>` after a parameter list containing at least one **hard** modifier (`ref`/`out`/`in`/`readonly`). Because the token appears in exactly one grammar rule, the parser commits to the lambda interpretation unambiguously.

`scoped` is intentionally NOT a hard modifier in the scanner: the existing *Scoped contextual keyword* test pins `(scoped p) => null` as `parameter_list(parameter(type=identifier(scoped), name=p))`, and that shape is preserved. This is also more spec-aligned — the C# 14 spec treats `scoped` as auxiliary, combining with `ref`/`ref readonly`.

## Test plan

- [x] `tree-sitter generate` succeeds with no new warnings or "unnecessary conflict" notices
- [x] `tree-sitter test` — 183 passing, 0 failing (baseline was 168; this PR adds 15 new corpus tests across 7 features)
- [x] Negative sentinels confirm the scanner does NOT over-fire: `(ref obj);` continues to parse as `parenthesized_expression(ref_expression(...))`, `(ref x) + 1` continues to parse via the pre-existing `cast_expression` path, and `(ref int i) => i` continues to parse as `parameter(type=ref_type(...), name=i)` (pinned by the existing *Lambda expression with ref modifier* test)
- [x] Method-declaration parameter lists with parameter modifiers (`void M(out int a)`, `void M(scoped ref scoped p)`) continue to parse exactly as before — verified by diff against pre-change baseline

## Supersedes

This PR consolidates the six smaller PRs (#423–#428) opened earlier. Those were closed when the feature branches were rebased into this linear history.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,10 @@ C# grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter) based u
 
 ### Status
 
-Comprehensive supports C# 1 through 13.0 with the following exception:
+Comprehensive supports C# 1 through 14.0 with the following exceptions:
 
 - [ ] `async`, `var` and `await` cannot be used as identifiers everywhere they are valid
+- [ ] File-based apps preprocessor directives (`#:property`, `#:package`, `#:sdk`, `#:project`) are not yet recognized
 
 ### References
 
--- a/grammar.js
+++ b/grammar.js
@@ -57,6 +57,10 @@ export default grammar({
 
     [$.lvalue_expression, $._name],
     [$.parameter, $.lvalue_expression],
+    // C# 14 null-conditional assignment: `obj?.x = v` requires
+    // `conditional_access_expression` to appear in both lvalue and
+    // non-lvalue contexts. Tree-sitter needs GLR for the decision.
+    [$.lvalue_expression, $.non_lvalue_expression],
 
     [$.type, $.attribute],
     [$.type, $.nullable_type],
@@ -1395,6 +1399,9 @@ export default grammar({
       alias($.bracketed_argument_list, $.element_binding_expression),
       alias($._pointer_indirection_expression, $.prefix_unary_expression),
       alias($._parenthesized_lvalue_expression, $.parenthesized_expression),
+      // C# 14 null-conditional assignment: `obj?.x = v`, `obj?[i] = v`.
+      // https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#null-conditional-assignment
+      $.conditional_access_expression,
     ),
 
     // Covers error CS0201: Only assignment, call, increment, decrement, await, and new object expressions can be used as a statement
--- a/grammar.js
+++ b/grammar.js
@@ -109,6 +109,12 @@ export default grammar({
     [$.using_directive],
 
     [$._constructor_declaration_initializer, $._simple_name],
+
+    // For C# 14 extension declarations: `extension(scoped X x)` —
+    // `scoped` can be a parameter modifier, a scoped_type, or a
+    // reserved identifier (when there's no further type). Keep all
+    // three alive until later tokens disambiguate.
+    [$.receiver_parameter, $.scoped_type, $._reserved_identifier],
   ],
 
   externals: $ => [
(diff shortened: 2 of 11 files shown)


AlexLaroche:
@tris203 @damieng

damieng:
Hi Alex, thaniks for the PR!

It looks like there's 3 files not parsing in the .cs - it might be a regression on your rules or it could be they're doing stuff upstream on the files. I think we're good either way you can just add the excluded files to the parse examples exclusions in https://github.com/tree-sitter/tree-sitter-c-sharp/blob/master/.github/workflows/ci.yml and we can get this merged in then rebase the other one and merge that in too.

Nice work!

AlexLaroche:
## CI failure analysis + fix

The three `Test parser` failures on the previous run all manifested in the `Parse examples` step. After investigating, only **one** was actually caused by this PR; the other two were latent gaps on `master` that became visible because `master`'s CI workflow has not actually run since 2026-04-14 (path-filtered `push` trigger, no `schedule:`) — so 49 days of upstream PowerToys / NUnit corpus drift went unvalidated until this PR re-triggered the example parse on today's `main`.

### What failed and why

| File | Cause | Pre-existing on `master`? |
|---|---|---|
| `examples/PowerToys/.../ImplicitWildcardQueryBuilder.cs` | List-pattern slice with a bound subpattern: `['"', .. var inner, '"']`. `list_pattern` accepted bare `..` but not `'..'` followed by a `pattern` — the C# 11 slice-pattern form. | ✅ pre-existing |
| `examples/nunit/.../TypeExtensionTests.cs` | Collection expression `[type]` inside an object initializer (`{ TypeArgs = [type] }`). `type` (and 7 other `attribute_target_specifier` keywords — `field`, `method`, `param`, `property`, `return`, `event`, `typevar`) won the LR resolution over `_identifier_token` despite `word: $._identifier_token` being set, so the element parsed as ERROR. Latent since collection-expression support landed in #402 (April 2026). | ✅ pre-existing |
| `examples/PowerToys/.../SettingsBackupAndRestoreUtils.cs` | **Actually introduced by this PR**, specifically by the lambda-modifier external scanner in `7c7b5f3`. `scan_lambda_paren_open` advanced forward speculatively past `(` chasing a possible `(ref x) => ...`, hit `$` and bailed with `return false`, but the lexer cursor was now past `(false, $` and the wrapper fell through to the rest of `tree_sitter_c_sharp_external_scanner_scan`. The next condition that fired — `INTERPOLATION_REGULAR_START` — emitted a token whose span started at the original scan position and ended at the dirty cursor, swallowing `(false, $"` and cascading the rest of the tuple into ERROR. Minimal repro: `return (false, $"Invalid {p}", "Error");` | ❌ this PR |

### How they were fixed

This PR now stacks on `fix/upstream-corpus-gaps` (3 commits, also opened separately as an upstream-corpus-gap fix branch):

- **`d8e28bb`** — adds `slice_pattern` (`'..' optional($.pattern)`) inside `list_pattern`; adds the 6 non-keyword attribute-target literals to `_reserved_identifier`; bumps `attribute_target_specifier` to `prec(1)` so `[target:` still parses as an attribute target when `:` follows. Fixes #1 and #2.
- **`68a0df9`** — null-conditional assignment (the original `d63b68d` from this PR, preserved through the rebase).
- **`d359ca8`** *(new on this branch)* — changes `scan_lambda_paren_open` to a tri-state return (`SUCCESS` / `NO_PAREN` / `FAILED_AFTER_PAREN`) so the wrapper can distinguish "didn't start, safe to fall through" from "started and failed, must return false so tree-sitter rewinds." Fixes #3 without affecting the lambda-parameter success path.

### Local validation

- `tree-sitter test`: **184 / 184** corpus tests pass.
- Full local replay of the `Parse examples` step (`examples/Newtonsoft.Json`, `examples/nunit`, `examples/orchard`, `examples/PowerToys` at upstream HEAD): **9173 / 9173 files parse**.
- Master parses 9170 / 9173 of the same set today, with the 3 failures listed above. This branch fixes all 3.

AlexLaroche:
@damieng could you take a look when you have a moment? See the analysis above for the failure breakdown and the fix layout.

damieng:
Thanks Alex!

PR #439, opened 2026-08-06
test: add failing corpus cases for eight known parse gaps

alexdarch:
## ⚠️ CI will be red — that is the point of this PR

This adds **8 corpus tests that fail today**. No grammar or scanner change is
included. Each expected tree is the tree a *correct* parse should produce, so
the suite goes from 0 to 8 failures. It's a bug report expressed as executable
tests, not something to merge as-is — happy to reshape it into whatever form is
useful (issues, draft-until-fixed, or paired with fixes in follow-ups).

`src/parser.c` and `src/grammar.json` are untouched, and I verified that
regenerating from the committed `grammar.js` reproduces both byte-for-byte, so
the "parser matches grammar" check is unaffected.

### `expressions.txt`

**`in`/`out` on a typed lambda parameter** — `(in int value) => value`

The parameter reading is shadowed by `argument`'s
`optional(choice('ref','out','in'))`, so the input reduces to
`argument → declaration_expression` and then errors on `)`. `--debug` shows no
GLR split (`version_count` stays 1). Only `in`/`out` are affected: `ref int x`
escapes because it also reads as `ref_type` (which is why the existing *Lambda
expression with ref modifier* test records `type: (ref_type …)` and no
`(modifier)` node), and `this`/`scoped`/`readonly` aren't argument prefixes.
Deleting `'in'` from `argument` in a scratch copy makes the input parse as
`(parameter (modifier) type: (predefined_type) name: (identifier))`.

**Assignment through a cast pointer** — `*((uint*) ptr) = 1`, `*(ulong*) ptr = 2`

`_pointer_indirection_expression` restricts `*`'s operand to an
`lvalue_expression`, but a cast — parenthesized or bare — is a non-lvalue, so
neither form parses. In C# `*` applies to any pointer-typed expression and
yields an lvalue. This looks like fallout from the `&`/`*` split in #430.

**Assignment to a ref-returning call** — `Slot() = value`

`assignment_expression`'s `left` is `lvalue_expression`, which has no
`invocation_expression`, so a `ref`-returning method can't be assigned through.

### `contextual-keywords.txt`

**`required` as an ordinary identifier**

`required` is in `modifier` but missing from `_reserved_identifier` (unlike
`file`, `scoped`, `var`, …). As a parameter name it's fine — keyword extraction
falls back to `_identifier_token` where the keyword isn't valid — but in
statement position `local_declaration_statement`'s `repeat($.modifier)` makes
the keyword valid, so `required = true;` commits to a local declaration and dies
at `=`. `required` isn't a legal local-variable modifier in C# at all. The
`[$._reserved_identifier, $.modifier]` conflict already exists, so adding
`'required'` to `_reserved_identifier` looks like the minimal fix.

### `literals.txt`

**Nested braces in an interpolated string** — `$"{{{string.Join(", ", value)}}}"`

Scanner bug. `brace_advanced` is a single function-scope variable shared by the
`INTERPOLATION_OPEN_BRACE` and `INTERPOLATION_STRING_CONTENT` branches. On
`{{{`, OPEN_BRACE consumes the first `{`, leaves `brace_advanced == 1`, then
bails because the next char is `{`. STRING_CONTENT inherits the stale `1` while
`open_brace_count == 0`, so its `brace_advanced == open_brace_count` stop can
never fire at the third brace — and it's only reset once a non-`{` is seen, by
which point all three braces are consumed. Content then runs to the first `"` it
meets, which is the one inside `", "`. Result: `string_content` = `{{{string.Join(`,
the interpolation never opens, and `", "` becomes a separate `string_literal`.
`$"{{x}}"` works because two braces never need the third-brace stop.

**Empty interpolated string** — `$""`

The scanner consumes both quotes probing for a raw string and, finding only two,
deliberately pushes nothing onto `interpolation_stack` ("If there's only two
quotes, that's just an empty string") — but still returns
`INTERPOLATION_REGULAR_START` with `mark_end` after the `$`. The grammar then
requires `interpolation_start_quote`, whose handler needs a non-empty stack and
returns false. Result: `(ERROR (interpolation_start)) (string_literal)`.

### `preprocessor.txt`

**`#if` around a base list**

No `preproc_if` variant is valid in base-list position — `preprocIf` is
instantiated for declarations, top-level items, expressions, enum members and
attribute lists only, and `_class_declaration_initializer` accepts just
`type_parameter_list | parameter_list | base_list`. A
`preprocIf('_in_base_list', $ => $.base_list, 0, false)` variant allowed in the
class/struct/interface/record initializers would cover it. **The expected tree
here is a guess at house style** (`preproc_if` as an unnamed child of
`class_declaration`, following the `preproc_if_in_attribute_list` precedent) —
no corpus test exercises that variant, so please redirect it if you'd shape the
node differently.

**`#if`-controlled braces inside a method body**

`preproc_if_in_top_level` content is `choice($._top_level_item_no_statement,
$.statement)`, and a lone `{` or `}` is neither. **This one is not fully
fixable**: the `{` sits in one `#if…#endif` region and its `}` in another, so a
node pairing them would have to overlap the two `preproc_if` nodes. The
expectation records the best a tree can represent — no ERROR, with the
conditional scope simply unmodelled — reachable only by letting `preproc_if`
content accept a bare brace. Include it or drop it as you prefer; it's a
limitation record more than a bug.

## Notes for anyone picking these up

- The `*`/lvalue and ref-return fixes both widen exactly the machinery #430
  narrowed to make `(a) && (b > 0)` parse correctly, so re-run the
  *Logical AND vs cast disambiguation* test and the `examples/**` sweep after
  either.
- This grammar has an external scanner, so use `tree-sitter test -r`. Without
  the forced rebuild a stale compiled artifact can invent failures — it cost me
  a wrong diagnosis while writing this up.


alexdarch:
@amaanq this was draft and clearly not finished. Closing it is a bit of a trigger happy response. I haven't been through all of the code manually yet, and Claude clearly has done some crap, but that doesn't mean that the test cases that I wrote are wrong - this tree-sitter is still producing ERROR ASTs for those cases and they ought to be fixed

amaanq:
1) Bugs should be reported in an issue, not thrown into an unactionable draft pr which makes zero sense
2) The entire pr, including prose, was llm generated. Why should I look at something you haven't looked at? I'm not against ai usage entirely but this is slop.

alexdarch:
I didn't say to look at it yet? you just came in 1hr after I made it and closed it. I'm happy to make a bug report, or whatever you want - just felt a but like you jumped the gun and are now getting upset about something that I hadn't even intended for anyone to see. The reason I made the PR is because I like reviewing changes in the github ui...

But regardless of that, what I've actually done is distilled down 8 minimal cases of where the current code produces ERROR nodes from a large codebase at my company. I'm happy to give this another go, or create 1 or 8 bug reports with them - whatever you prefer.

amaanq:
If you did not intend for anyone to see it, then you should not be opening a PR upstream, that is what your fork is for. You can make PRs with your branch in your own repo. No one is upset, this is just poor etiquette and is being treated as such. You're welcome to file one issue if the bugs are small/interlinked, or multiple if some are larger.

